PR #4, opened 2021-05-28
emit_signal("onMDAFinished", sequence) 

fdrgsp:


tlambert03:
if you want the signal to be omitted with the sequence, you'll need to add it here:
https://github.com/tlambert03/pymmcore-remote/blob/044ff0730949725bac97a167fa63b2f0ed71c9c5/pymmcore_remote/_mmcore_plus.py#L114

tlambert03:
great, now let's update the test in `_tests/test_client`...

update the signals and checks [here](https://github.com/tlambert03/pymmcore-remote/blob/044ff0730949725bac97a167fa63b2f0ed71c9c5/pymmcore_remote/_tests/test_client.py#L57-L58), with something like the following:

```python
    def _check_finished(obj):
        return obj.uid == mda.uid

    signals = [cb.MDAFrameReady, cb.MDAFrameReady, cb.MDAFinished]
    checks = [_test_signal, _test_signal, _check_finished]
```

tlambert03:
tests are failing because  we're not using the latest (unreleased) version  of useq-schema... can you update `setup.cfg` to replace the `useq-schema` entry with:

```ini
testing =
    useq-schema @ git+https://github.com/tlambert03/useq-schema.git
```

fdrgsp:
I've added the new testing useq-schema to the PR

tlambert03:
thanks!  remaining fail is unrelated.

PR #5, opened 2021-07-01
add typing_extensions to install requires

ianhi:
It's used here: https://github.com/tlambert03/pymmcore-remote/blob/25982bc9194f48247781b616b1b8c558b3eff32c/pymmcore_remote/_client.py#L9


ianhi:
oops also looks as though `useq` is currently also a hard requirement for import:
https://github.com/tlambert03/pymmcore-remote/blob/25982bc9194f48247781b616b1b8c558b3eff32c/pymmcore_remote/_serialize.py#L9

Should that be added to the requires? or put in as an optional import?

tlambert03:
thanks! 

> Should that be added to the requires? or put in as an optional import?

lets go ahead and add it.

PR #8, opened 2021-07-02
Rename to pymmcore-plus, update readme

tlambert03:
I think pymmcore-plus is a better name for this repo (since it's not strictly about remote usage).  also updating the readme.  will publish to pypi tomorrow


ianhi:
:+1:  good change. One follow up thought is that it feels like this repo is really providing two different functionalities: `pymmcore-plus` and `pymmcore-plus-remote`. The remote aspect is nice for integration with napari etc but also seems to make the error messages more confusing - whereas`pymmcore-plus` sans `pyro5` is also very useful and I imagine would have less obfuscated error messages.

Perhaps even could be nice to have multiple imports:

```python
from pymmcore-plus import MMCorePlus

# grab remote
from pymmcore-plus import RemoteMMCorePlus
```

ianhi:
oh! I see that is already possible - my bad

tlambert03:
will add that import to the readme!  Also, it _is_ possible to improve the remote tracebacks... just need to [twiddle with it](https://pyro5.readthedocs.io/en/latest/errors.html#remote-exceptions) a bit

PR #12, opened 2021-07-06
Add automatic micromanager discovery for linux

ianhi:
If you follow the steps in the Docker file for pymmcore (https://github.com/micro-manager/pymmcore/blob/main/Dockerfile)
then you end up with an installation in `/usr/local/lib/micro-manager/`

Also, paths are case senstive on linux hence so search for both capital
and lowercase `m` in the glob.

Finally, added the searched path to the logger if nothing was found in
case people have non-standard install locations.

---

Annoyingly however this won't automatically grab the demo config as the install instructions don't move it to `/usr/local/lib/micro-manager`  so you still need to explicitly locate the path

### DIFF
--- a/pymmcore_plus/_util.py
+++ b/pymmcore_plus/_util.py
@@ -26,10 +26,11 @@ def find_micromanager() -> Optional[str]:
     applications = {
         "darwin": Path("/Applications/"),
         "win32": Path("C:/Program Files/"),
+        "linux": Path("/usr/local/lib"),
     }
     try:
         app_path = applications[sys.platform]
-        pth = str(next(app_path.glob("Micro-Manager*")))
+        pth = str(next(app_path.glob("[m,M]icro-[m,M]anager*")))
         logger.debug(f"using MM path found in applications: {pth}")
         return pth
     except KeyError:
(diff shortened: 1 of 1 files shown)


tlambert03 (approved):
awesome thanks!  

are you actually using this on linux out of curiosity?  Or just saw the `NotImplementedError` and filled it in?

tlambert03:
as an aside ... it would be kinda cool if we could get something like that docker runfile into the [installer script](https://github.com/tlambert03/pymmcore-plus/blob/main/pymmcore_plus/install.py) for convenience.

ianhi:
> are you actually using this on linux out of curiosity? Or just saw the NotImplementedError and filled it in?

Yup! My personal computer is linux only so I ran into this pretty much immediately.

ianhi:
> as an aside ... it would be kinda cool if we could get something like that docker runfile into the installer script for convenience.

Indeed. Also maybe a `linux-troubleshooting.md` could be helpful?

tlambert03:
cool.  do you know how many of the device adapters would work on linux?  or mostly just convenient for development? (similar to mac i guess)

ianhi:
I don't know because I haven't really had the opportunity to try (I only really have access to three micrscopes all of which have setups that I'd prefer not accidently undo their working status. Plus mostly they are not primarily in my care). I got this going so I could mess around with this repo.

PR #14, opened 2021-07-07
Unify callback API and restructure package

tlambert03:
this PR aims to cleanup the callback API a little bit, so that one can simple swap out the local `CMMCorePlus` object for the `RemoteMMCore` without changing any syntax.

for instance:
```python
from pymmcore_plus import CMMCorePlus, RemoteMMCore

# local
mmc = CMMCorePlus()
mmc.systemConfigurationLoaded.connect(lambda: print("loaded"))
mmc.loadSystemConfiguration("demo")

# remote
with RemoteMMCore() as remote_core:
    # no need to create and register a new callback object
    remote_core.systemConfigurationLoaded.connect(lambda: print("loaded"))
    remote_core.loadSystemConfiguration("demo")
```


tlambert03:
@ianhi, curious to hear your thoughts on this.  don't need to review the code if you don't have time, mostly curious to hear if you think the API is better

pymmcore_plus/_client.py
```diff
@@ -39,10 +54,34 @@ def __init__(
...
+        connect to any of the signals in the `_CMMCoreSignaler` class:
+
+            proxy = RemoteMMCore()
+            proxy.systemConfigurationLoaded.connect(lambda: print("loaded!"))
+
+        This method can be used to register an independent callback object.
+
+        Parameters
+        ----------
+        callback : CallbackProtocol
```
ianhi, line 75: It's a bit confusing to use this in docstring as this isn't a public object that people can normally interact with or see documented. Maybe it should be imported in `__init__`?

pymmcore_plus/_client.py
```diff
@@ -39,10 +54,34 @@ def __init__(
...
+        self.register_callback(_CBrelay(self))
 
     def register_callback(self, callback: CallbackProtocol):
+        """Register callback object in proxy process to receive remote callbacks.
+
+        Not to be confused with `mmcore.registerCallback`, which is only used once by
+        CMMCorePlus to receive events coming from the internal C++ CMMcore object.
+
+        Note: RemoteMMCore automatically behaves as a callback receiver, you can
+        connect to any of the signals in the `_CMMCoreSignaler` class:
```
ianhi, line 66: This sentence threw me for a loop on the first read. I think this is the first mention of `callback receiver`. 

I was going to suggest a clarification but on further reading I'm struggling to see the use case here. Is there a good reason for this to be public now that you can do `mmcore.<signal>.connect` for all of the possible signals?

ianhi (approved):
:+1:. The only other thing I might consider is hiding the `register_callback` as private as in the comment.

I also think an example with more expository text will eventually be helpful. I'm happy to write one in a follow up to this which could carry the advantage of me testing my understanding here and thus also act as a check that all the ideas are clear. (Though this would be next week - going on vacation tomorrow!)

pymmcore_plus/_client.py
tlambert03: > Is there a good reason for this to be public now that you can do mmcore.<signal>.connect for all of the possible signals?

well, I can't really think of one! :joy: ... so I'll make it private

pymmcore_plus/_client.py
tlambert03: for exactly that reason, I tend not to worry too much about custom types in numpy docstrings.  at this point in the project having the class above  `class CallbackProtocol(Protocol):` feels like sufficient documentation to me
(I'd much prefer a world in which we all moved to using type annotations in signatures, and leaving them mostly out of docstrings).   exporting it in the main namespace is sort of an independent question (and if we're making this private, then i probably wouldn't want to do that). the one thing we could do (which is ugly, but accurate) would be to change it to `callback : pymmcore_plus._client.CallbackProtocol` ... but for reasons mentioned above, I'd probably skip that too.

tlambert03:
> I also think an example with more expository text will eventually be helpful. I'm happy to write one in a follow up to this which could carry the advantage of me testing my understanding here and thus also act as a check that all the ideas are clear.

sure that would be great.  The example in `qt_integration` covers most of what I think needs to be understood (along with a more general understanding of callbacks, describe in the [psygnal readme,](https://github.com/tlambert03/psygnal) and beyond the scope here).

note: the only "Qt" thing about the `qt_integration` example at this point is the event loop that waits for the `mmcore.sequenceFinished` signal that triggers `app.quit` (otherwise, the process running the remote proxy would simply quit), and the button that lets you quit early.  I think it would be great to have an example of a pure-python [asyncio version](https://docs.python.org/3/library/asyncio-eventloop.html) too.

pymmcore_plus/_client.py
tlambert03: also, it says on the next line: "Just an object that has a `receive_core_callback` method"

tlambert03:
one more note:  I moved things into submodules `core`, `client`, `server` ... so if you have any open branches, you'll probably want to merge master and resolve conflicts sooner than later (sorry about that ... but it's probably better in the long run)

PR #19, opened 2021-07-14
bring back qcallback

tlambert03:
This brings back the QCoreCallback object removed in #14 ... which was the reason why napari-micromanager was working (when directly calling core.methods in a callback).

I'm still not 100% sure why the Qt callback works but a synchronous callback using psygnal doesn't (meaning it makes sense why the synchronous psygnal callback fails, but I'm not sure what it is about the QObject that "works").

In any case, it deserves further investigation, but this at least gets us up and running again.

By default the proxy will detect whether its running in a Qt environment, and use the QCoreCallback instead of the synchronous version.  (but Qt is not required to run the synchronous version in a local setting)

this also puts the callbacks into the `core.events` namespace (rather than at the top level).

also, removing the typing from this repo, since I [added them upstream](https://github.com/micro-manager/pymmcore/pull/46).


ianhi:
Perhaps something of an explanation can be found in https://doc.qt.io/qt-5/signalsandslots.html

> When a signal is emitted, the slots connected to it are usually executed immediately, just like a normal function call. When this happens, the signals and slots mechanism is totally independent of any GUI event loop. Execution of the code following the emit statement will occur once all slots have returned. The situation is slightly different when using queued connections; in such a case, the code following the emit keyword will continue immediately, and the slots will be executed later.

Does this count as a `queued connection`?

tlambert03:
Yeah that's my assumption.   as you'll see here, the default is for direct connection unless the receiver lives in another thread: https://doc.qt.io/qt-5/qt.html#ConnectionType-enum

In this case I had been assuming that the callback would have been called in the client thread.  But I suspect that the thread that I create in the client code is making this "work" in qt for exactly this reason.  (we could test it by specifying directConnection when we connect the callback)

If so, it would be cool to figure out a similar "queued connection" mechanism for psygnal

tlambert03:
yeah, fascinating:  If I change it to
```
core.events.systemConfigurationLoaded.connect(_cb, Qt.DirectConnection)
```

then the Qt test fails with:

```
Pyro5.errors.PyroError: the calling thread is not the owner of this proxy, create a new proxy in this thread or transfer ownership.
```
which is explained [here](https://pyro5.readthedocs.io/en/latest/clientcode.html#proxies-connections-threads-and-cleaning-up) (And is why I have `_pyroClaimOwnership` in the server-side `emit` method).

incidentally, that's the same error I get in the psygnal-based callback if I emit the client-side signal [with `asynchronous=True`](https://github.com/tlambert03/psygnal#emitting-signals-asynchronously-threading).  So, i think you're definitely right that the Queued connection here is the key.  Would be a fun exercise to add that to psygnal.  Not urgent, since our immediate need for running the core in another process is so that we can run a qt event loop in the main process.

tlambert03:
(for my own future reference, the C code that determines how a callback should be connected in Qt starts [around here](https://code.woboq.org/qt5/qtbase/src/corelib/kernel/qobject.cpp.html#3755))

PR #24, opened 2021-07-16
Set exposure after setting channel in run_mda

ianhi:
Depending on user configuration setting the channel can have a side effect of changing the exposure. This then overrides the event exposure. Since the event is a more local config than the channel config it should take precedence for setting the exposure.


tlambert03 (approved):
works for me!  thanks

PR #29, opened 2021-07-29
First pass at MDA writers

ianhi:
First writer is a little rough but wanted something to play with in order to explore this API.

I also added a demo config to examples folder for two reasons: 1. It's helpful on linux. 2. It gives a consistent demo example across installations.


pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -157,7 +158,13 @@ def getCameraChannelNames(self) -> Tuple[str, ...]:
...
         )
 
-    def run_mda(self, sequence: MDASequence) -> None:
+    def run_mda(self, sequence: MDASequence, writer: MDAWriter = None) -> None:
+
+        if writer is not None:
+            for e in dir(self.events):
+                if "sequence" in e or e == "frameReady":
+                    getattr(self.events, e).connect(getattr(writer, e))
+
```
ianhi, line 167: I ended up using the events rather than specifically calling the methods of the writer. This felt cleaner because otherwise you are constantly checking `if writer is not None`. Although I suppose alternatively we could start with:
```python
if writer is None:
    writer = DummyWriter()
````

File: pymmcore_plus/mda_writers/_writers.py [hunk 2]
```diff
@@ -0,0 +1,65 @@
...
+
+    def sequencePauseToggled(self, paused: bool) -> None:
+        ...
+
+    @staticmethod
+    def get_axis_order(seq: MDASequence) -> Tuple[str]:
+        """Get the axis order using only axes that are present in events."""
+        event = next(seq.iter_events())
+        event_axes = list(event.index.keys())
+        return tuple(a for a in seq.axis_order if a in event_axes)
```
ianhi, line 35: This is something I found necessary as you can run into errors when querying an even for it's index if the axis order contains a dimension that isn't actually used to create an event. e.g.:

```python
mda = MDASequence(
    # stage_positions=[(100, 100, 30), (200, 150, 35)],
    channels=["DAPI", "FITC"],
    time_plan={'interval': 1, 'loops': 20},
    z_plan={"range": 4, "step": 0.5},
    axis_order='tpcz',
)
```

tlambert03: yeah, I've run into this a lot as well.  it's annoying.  We should think about how to fix this upstream as well.

pymmcore_plus/core/_mmcore_plus.py
tlambert03: i like this pattern... it's actually pretty general though, so I wonder if perhaps we should make a context manager that lets you temporarily connect an "event listener"?

```py
# or core.events_listener():
with core.events.listener(MDAWriter()):
    core.run_mda(seq)
```

tlambert03 (commented):
yeah, I think I like it!  will have a closer look later, left a few comments

tlambert03:
a thought:

rather than a general event follower: do you think we could narrow down the scope of the writer protocol?  It seems like it probably needs three methods: 
- initialize: provide knowledge of the axis names involved, and the expected `shape` of the final data
- add_frame: provide a new 2D image, and slice object corresponding to it's position in the `shape` of the full data
- finalize: let it know it's done.

one thing to think about there is how it would handle an acquisition with unknown shape (like, acquire until some event happens)... but that simple protocol would satisfy common usage, and simplify what the writer itself needs to extract from the sequence/event objects?

ianhi:
That sounds great to me, it will make them much simpler to document how to create and remove some boilerplate that was going to end up in every writer. The one caveat I'd make is that `finalize` should probably allow for receiving whether the acquisition finished completely or was stopped prematurely. 

I also imagine that this would not lead to eliminating the existing MDA events? With those in place if someone wants to make a more complex mda writer then they still can of course.

Also, the event listener context manager you noted above is independently a good idea

tlambert03:
> I also imagine that this would not lead to eliminating the existing MDA events? With those in place if someone wants to make a more complex mda writer then they still can of course.

yeah, I think it's probably best to provide the `MDAEvent` object as well

ianhi:
```python
class MDAWriter(Protocol):
    def initialize(self, shape, axis_order, seq: MDASequence) -> None: ...
    def addFrame(self, image, slice, event: MDAEvent) -> None: ...
    def finalize(self) -> None: ...
```

So something like the above? Leaving in the underlying events and sequence seems valuable as it provides extra flexibilty

ianhi:
May also be nice to include expected dtype. Is there an easy way to access that without snapping an image?

ianhi:
ahh, it's already done in the state. I will integrate that.

ianhi:
working on expanding the tests

ianhi:
Only failure is codecov which seems to be because the base protocol isn't exercised in the tests and there isn't a test without `tifffile` installed:
image


@tlambert03 I think I'm happy with this for now so whenever you have time to look at it that'd be great.

( I guess the one improvement is that the tifffile writer should accept `pathlib.Path` objects as well)

examples/run_mda.py
```diff
@@ -11,5 +11,8 @@
 )
 
 mmc = CMMCorePlus()
-mmc.loadSystemConfiguration("demo")
-mmc.run_mda(sequence)
+mmc.loadSystemConfiguration("demo_config.cfg")
```
tlambert03, line 14: this now only works if you cd into the `examples` directory first (I tend to run `python examples/whatever.py`)... can you make it `str(Path(__file__).parent / "demo_config.cfg")`

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -283,9 +285,18 @@ def getCameraChannelNames(self) -> Tuple[str, ...]:
             for i in range(self.getNumberOfCameraChannels())
         )
 
-    def run_mda(self, sequence: MDASequence) -> None:
+    def run_mda(self, sequence: MDASequence, writer: MDAWriter = None) -> None:
```
tlambert03, line 288: I'm not in love with having run_mda take a new parameter, i can imagine a writer being just one of many things that may want to peek in on the running experiment and do something with it.  This also makes it hard to add, two writers (who knows why you'd want to do it, but it's not a ridiculous concept...).

I still think I'd prefer the concept of a [temporary signal listener](https://github.com/tlambert03/pymmcore-plus/pull/29#discussion_r679283108)
any thoughts there?

I also think we should just improve the `axis_order` getting on the `MDASequence` object as needed, rather than adding convenience methods here (`get_axis_order`) and then passing `sequence.shape`, `get_axis_order(sequence)`, _and_ `sequence` to the writer... when it seems like we could just pass `sequence`.

ianhi: > I'm not in love with having run_mda take a new parameter, 

That's sensible but I don't quite see how that would work with your earlier suggestion about narrowed scope (unless the the temporary listener did some magic to find the correct methods).:

> rather than a general event follower: do you think we could narrow down the scope of the writer protocol? It seems like it probably needs three methods:

    initialize: provide knowledge of the axis names involved, and the expected shape of the final data
    add_frame: provide a new 2D image, and slice object corresponding to it's position in the shape of the full data
    finalize: let it know it's done.


> I also think we should just improve the axis_order getting on the MDASequence object as needed, rather than adding convenience methods here (get_axis_order) and then passing sequence.shape, get_axis_order(sequence), and sequence to the writer... when it seems like we could just pass sequence.

Agreed, I'll try to open a PR on useq

dpshepherd:
Hi all-

First off, thank you so much your work on magic-gui, pymmcore-plus, and Napari!!!

Following up here based on Talley's suggestion in the twitter-verse (https://twitter.com/TalleyJLambert/status/1466840754404433921), I wanted to drop in here now that we have working acquisition and writing code. I'm 100% positive we are not doing things "correctly" yet, and I am happy to learn on how to become correct on the threading & signaling side of things beyond this file saving topic.

Caveats aside, we (mainly me) recently implemented hardware triggered control of our oblique plane microscope using Napari and pymmcore-plus. We have both [high-speed multi-color timelapse](https://github.com/QI2lab/OPM/blob/7933eabf8616a6808914bb91d29b598b1e864d0f/napari-control/opm_timelapse_control.py) or [constant speed stage-scanning using fluidics](https://github.com/QI2lab/OPM/blob/7933eabf8616a6808914bb91d29b598b1e864d0f/napari-control/opm_iterative_control.py) experiments in the lab. We use the latter for MERFISH, but it will work for anything that requires iterative work (iterative FISH, CODEX, _in situ_ sequencing, etc...). It should be easily adaptable to a standard microscope that acquire independent tiles instead of a constant speed stage scan.

For both use cases, we ended up using a Zarr backed filestore without compression. We are keeping the metadata in a separate .CSV files, because it seems like OME-NGFF is in flux and also doesn't have a clear specification for timelapse and/or highly multiplexed data yet.

One immediate project (beyond fixing all the hacky code) that comes to mind is that we adopt [useq-schema](https://github.com/tlambert03/useq-schema) and start discussing ways to store metadata and data correctly in the Zarr. I think the multi-color timelapse would be a better place to start with this?

Doug

tlambert03:
thanks for your comments @dpshepherd!  Super glad to hear you're still banging away at this, and would love to chat more about it.  

With regards to useq-schema, I'm really glad to hear that you're thinking about adopting it.  I actually think that project has a lot of promise, and, if done well (as a small, well-defined but generalizable specification), has the potential to outlive all of these actual acquisition-engine projects.  I'll be the first to point out that it's _not_ currently done all that well 😂, as there are some hard assumptions about axes/dimensions in there that I'd very much like to abstract (while also keeping the "simple stuff simple")

I'll mention that @kasasxav (Xavier) has also expressed interest in generalizing `useq-schema` for his needs in [Imswitch](https://github.com/kasasxav/ImSwitch).  I suspect you two have non-overlapping, but similarly demanding experiment definitions.  Would you guys maybe be up for a zoom chat after the new year?

kasasxav:
Hi @tlambert03 and @dpshepherd, thank you for tagging me!

I think it would be great to meet and discuss next year about useq-schema, I'm quite excited about it and I think it would be nice if it works with different control software options. Daniel Fürth (@tractatus) also had some nice ideas about how to define experiments, so we can maybe meet altogether 👍 

I come back the 4th January, after that, I'm quite flexible!

tractatus:
@kasasxav @dpshepherd @tlambert03 sounds awesome. I like what I hear. I’m very excited about the possibility to define multichannel, multi-acquisition cycles experiments in compact reproducible useq-schema.

Currently in Texas visiting my sister but will be back and ready for meeting after the 4th like @kasasxav.

Let me know if you are all aboard a Zoom meeting and I can send out a Doodle.

dpshepherd:
@tlambert03 @kasasxav @tractatus - sounds good. I have a new postdoc who will work on these experiments joining the lab around the 4th (assuming he can fly from Europe...), so that timing works out perfectly on our end.

I agree having a univerisal-ish, compact definition for cyclic n-D datasets would be great. There is renewed movement across both the CZI cell atlas and BRAIN initiative for "better" and more general metadata collection on spatial -omics type experiments.

tractatus:
@dpshepherd sounds awesome! I've sent a doodle.

ianhi:
If this zoom hasn't happened yet could I join as well? very curious about following all of this.

tractatus:
@ianhi Absolutely no problem. Sorry about that. I sent you an invite over email. Let me know if the date and time would work for you.

ianhi:
Probably about time to revisit this now that we have the new `mda` object and the `register_mda_engine` method. 

I think we should aim to:

1. Make it easy for a user to use a  more complex writer of their choosing
2. provides at least a super basic saving mechanism (e.g. separate tiffs)

Some options for API

### 1. writers are responsible for registering their own callbacks and listening to `mdaEngineRegistered`


writers would follow the pattern of:

```python
class BaseWriter:
    def __init__(self, folder, prefix, other_metdata_etc, core: CMMCorePlus=None):
        self.core  = core or CMMCorePlus.instance()
        self.core.events.mdaEngineRegistered.connect(self._on_engine_registered)
    def _on_engine_registered(self, newEngine, oldEngine):
        """Register frameReady and sequence lifecycle signals"""
        ...
```

then usage would look like:

```python
# ... various setup commands
writer = TiffWriter(blah blah blah, core=core)

core.run_mda()
# when done disconnect if you want
writer.disconnect()
```

This is nice and simple and matches up with the pattern we've been using in napari-micromanager. It's also what I've been doing in https://github.com/ianhi/pymmcore-MDA-writers

### 2. Define writer protocol - register with either core or engine.

E.g. the protocol currently in this PR:
```python

class MDAWriter(Protocol):
    def initialize(self, shape, axis_order, seq: MDASequence, dtype: np.dtype) -> None:
        ...

    def addFrame(self, image, index, event: MDAEvent) -> None:
        ...

    def finalize(self) -> None:
        ...
```

and usage would look like:

```python
writer = TiffWriter(blah blah blah)
core.mda.register_writer(writer)

core.run_mda(seq)

# when done with that writer
core.mda.unregister(writer)


# swapping acquisition engines doesn't transfer writers
```

and then in engines would responsible for dispatching the correct methods in their `run` loop.


---
Curious to hear people's thoughts

tlambert03:
My first thought (and probably one reason this stalled a bit before) is that the protocol idea (`init`/`add_frame`/`finalize`) is _so_ similar to the existing signals `onMDAStarted`/`onFrameReady`/`onMDAFinished` that it feels to me a bit like an unnecessary additional layer.

The majority of the complexity here is really up to the writer, in how it handles a new frame.  it kinda seems like those three signals are sufficient for the vast majority of cases, and `my_writer = Writer(mda=core.mda)` would already work as is?

ianhi:
> The majority of the complexity here is really up to the writer, in how it handles a new frame. it kinda seems like those three signals are sufficient for the vast majority of cases, and my_writer = Writer(mda=core.mda) would already work as is?

It does indeed. I think that's more or less what I was going for in option 1. So then there are two questions I think.

1. Do we include a basic writer (and base class) here or are they fully delegated to another package
2. If the answer to 1 is yes then should writers take `mda` or should they take a `core` and then listen to the `mdaEngineRegistered` signal?

For 1. if included here then I think we could basically use the base class that I made here: https://github.com/ianhi/pymmcore-MDA-writers/blob/ac07f3ecd446adbb074b239d37c95c986752b90f/pymmcore_mda_writers/_writers.py#L25-L102

ianhi:
>  ome tiff writer

:+1: I stayed away from at first here those because I was burnt by micro-manager ome-tiff saving and how difficult (broadly impossible) it was to open them in python. But I think if we do the writing with `tifffile` then we should be garunteed to be able to read it with tifffile. This will require some digging around in tifffile to learn how to write them though.

tlambert03:
yeah. My (rather limited) understanding there is that the difficulties there had more to do with the actual Java code in micromanager that wrote the file.  which we wouldn't be re-using

ianhi:
> That said, we can help with the OME-TIFF writing using tifffile if needs be.

@dpshepherd even an a non-minimal example of how to correctly use tifffile would be really helpful for me!

dpshepherd:
> @dpshepherd even an a non-minimal example of how to correctly use tifffile would be really helpful for me!

Is the goal to append individual images to one large OME-TIFF as they are acquired? To do this using for tifffile + OME-TIFF the suggested way to is pre-create a blank OME-TIFF with the correct axes and then write into it using memmap. I have a couple meetings this morning, but I have this [post](https://forum.image.sc/t/how-to-create-an-image-series-ome-tiff-from-python/42730/7) bookmarked as one way to do it.

dpshepherd:
if you create the tif with `bigtiff=True`, this avoids the 4 gb limit.

ianhi:
> Would be great to have both options.... don't always love the massive single file (but good to have that link)

I may be confused here. I thought `ome.tiff` implied that the images were in a single stack file. Or at least that's the sense I got from micro-manager. Is it instead the case that stack vs many tiffs is orthogonal to whether or not it's an ome tiff?

tlambert03:
> I may be confused here. I thought ome.tiff implied that the images were in a single stack file.

that's one way, yes... but if I'm not mistaken the ome model also supports spreading dimensions across multiple files.  See https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#TiffData_UUID ... @joshmoore might be able to point us to a simple example of XML for a multi-file ome.tiff.  (I think the ASI dispim plugin in micromanager actually does that, will check)

joshmoore:
> @joshmoore might be able to point us to a simple example of XML for a multi-file ome.tiff

Sure. https://downloads.openmicroscopy.org/images/OME-TIFF/2016-06/companion/

> Is it instead the case that stack vs many tiffs is orthogonal to whether or not it's an ome tiff?

Yes. Being an OME-TIFF is really a matter of having OME metadata. And OME Metadata can turn multiple 2D TIFF arrays, whether in one or multiple files, into a single coherent 5D structure.

tlambert03:
thank you @joshmoore! 🙏

tlambert03:
thanks @ianhi  for all the great thinking here.  the library has evolved a fair amount since this, so I'm going to close this PR.  Feel free to reopen if you want to pick it up again

tlambert03:
this is superceded by #263, #267 and #265

PR #32, opened 2021-08-02
ENH: Check for mda canceled while waiting

ianhi:
Otherwise you can end up waiting for as long as the interval in the
sequence is (easily 10 or 15 minutes)

The test fails locally without the fix and passes with.  Using `time.sleep` instead of `waitSignals` I wasn't able get the ordering of events to work properly - however I'm not sure I'm actually using qtbot correctly here.

### DIFF
--- a/pymmcore_plus/core/_mmcore_plus.py
+++ b/pymmcore_plus/core/_mmcore_plus.py
@@ -163,22 +163,47 @@ class CMMCorePlus(pymmcore.CMMCore):
         self._paused = False
         paused_time = 0.0
         t0 = time.perf_counter()  # reference time, in seconds
-        for event in sequence:
-            while self._paused and not self._canceled:
-                paused_time += 0.1  # fixme: be more precise
-                time.sleep(0.1)
+
+        def check_canceled():
             if self._canceled:
                 logger.warning("MDA Canceled: {}", sequence)
                 self.events.sequenceCanceled.emit(sequence)
                 self._canceled = False
+                return True
+            return False
+
+        for event in sequence:
+            while self._paused and not self._canceled:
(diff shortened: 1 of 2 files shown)


tlambert03:
thanks @ianhi!  sorry for the delay.  I just modified the test here slightly and will merge now  instead of the perftimer, I used:

```py
mda = MDASequence(time_plan={"interval": 5000, "loops": 3})  # set long interval
with qtbot.waitSignals([proxy.events.sequenceStarted, proxy.events.frameReady]):
    proxy.run_mda(mda)
# set short timeout...
with qtbot.waitSignals([proxy.events.sequenceFinished, proxy.events.sequenceCanceled], timeout=500):
    proxy.cancel()
```

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -289,22 +289,47 @@ def run_mda(self, sequence: MDASequence) -> None:
...
+                    if to_go > 0.5:
+                        time.sleep(0.5)
+                    else:
+                        time.sleep(to_go)
+                    to_go = go_at - (time.perf_counter() - t0)
+
+            # check canceled again in case it was canceled
+            # during the waiting loop
+            if check_canceled():
+                break
```
ianhi, line 331: looking at this again it sure seems like it should return `False` here because the first check cancelled sets `self._canceled` to false.

ianhi: *if it was cancelled while waiting

and then we won't actually break of outer loop

ianhi: oooooh. nevermind, apparently I thought of this already and it does work just fine

PR #38, opened 2021-08-06
add setProperty override that emits propertyChanged

tlambert03:
related to #28 ... this reimplements `setProperty` to guarantee that an event is emitted if a property is changed using pymmcore.  It can't guarantee that an event will be emitted if it happens internal to the device adapter (and if the device adapter doesn't emit it), but it's a bit better.

@ianhi... now that I understand the issue slightly better what would you think about removing the `configSet` signal that we recently added and instead just reusing the `configGroupChanged` signal?  using a similar pattern to what I did in this PR:

```py
with self.events.configGroupChanged.blocked():
        super().setConfig(groupName, configName)
self.events.configGroupChanged.emit(groupName, configName)
```

is there a specific case where we wanted the native signal and the pymmcore-plus signal to have different names?

### DIFF
--- a/pymmcore_plus/core/_mmcore_plus.py
+++ b/pymmcore_plus/core/_mmcore_plus.py
@@ -45,6 +45,40 @@ class CMMCorePlus(pymmcore.CMMCore):
 
     # Re-implemented methods from the CMMCore API
 
+    def setProperty(
+        self, label: str, propName: str, propValue: Union[bool, float, int, str]
+    ) -> None:
+        """setProperty with more reliable event emission.
+
+        As stated by Nico: "Callbacks are mainly used to give devices the opportunity to
+        signal back to the UI."
+        https://forum.image.sc/t/micromanager-events-core-events-not-coming-through/53014/2
+
+        Because it's left to the device adapter to emit a signal, in many cases uses
+        `setProperty()` will NOT lead to a new `propertyChanged` event getting emitted.
+        But that makes it hard to create listeners (i.e. in the gui or elsewhere).
+
+        While this method override cannot completely solve that problem (core-internal
+        changes will still lack an associated event emission in many cases), it can at
+        least guarantee that if we use `CMMCorePlus.setProperty` to change the property,
+        then a `propertyChanged` event will be emitted if the value did indeed change.
+
+        Parameters
+        ----------
+        label : str
+            device label
+        propName : str
+            property name
+        propValue : Union[bool, float, int, str]
+            new value
+        """
+        before = super().getProperty(label, propName)
+        with self.events.propertyChanged.blocked():  # block the native event.
+            super().setProperty(label, propName, propValue)
+        after = super().getProperty(label, propName)
+        if before != after:
+            self.events.propertyChanged.emit(label, propName, after)
+
     def setDeviceAdapterSearchPaths(self, adapter_paths: ListOrTuple[str]):
         # add to PATH as well for dynamic dlls
         if (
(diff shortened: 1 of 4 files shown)


tlambert03:
ping @ianhi...  how do you feel about removing `configSet` and reusing `configGroupChanged`?

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -45,6 +45,40 @@ def __repr__(self) -> str:
...
+            device label
+        propName : str
+            property name
+        propValue : Union[bool, float, int, str]
+            new value
+        """
+        before = super().getProperty(label, propName)
+        with self.events.propertyChanged.blocked():  # block the native event.
+            super().setProperty(label, propName, propValue)
+        after = super().getProperty(label, propName)
```
ianhi, line 78: It's possible that this won't block the native event as I think there may be some async/threading happening where the property is set, a command goes out to the device (e.g. over serial) and some time later the device physically switches and emits a signal, and I don't think that mmcore always blocks on that. or at least once it goes into the device adapter I think all bets may be off as to what happens.

Though that's an edge case that liveable with

ianhi:
Woops sorry I missed this, took a long weekend starting on friday. 

---
> @ianhi... now that I understand the issue slightly better what would you think about removing the configSet signal that we recently added and instead just reusing the configGroupChanged signal? using a similar pattern to what I did in this PR:

This seems pretty reasonable. I do think that there is something a coherent difference between it being `set` and being `changed` (https://github.com/micro-manager/mmCoreAndDevices/issues/25#issuecomment-887683306) but it's simpler for for guis (napari-μmanager) to listen to just to existing signal and eventually use the new signal proposed here https://github.com/micro-manager/mmCoreAndDevices/issues/25#issuecomment-885746558. 

Blocking the signal like that does expose you to issues if there is a pathological settings file where two different config groups have overlapping settings (which I probably upsettingly common)  so that changing one will also result in a change in the other. As a sketch of what I mean take two groups that have some overlapping settings:

```
Group1:
     - Setting 1
     - Setting 2
     - Setting A
     - Setting B
 
Group2:
     - Setting 3
     - Setting 4
     - Setting A
     - Setting B
```

With that its possible that `setConfig("Group1", "preset_whatever")` should also trigger a `onConfigGroupChanged` for `Group2` which will be silenced by your proposed change. These were just quick thoughts as I'm busy for a bit today, I'll try to think about how bad of a thing that feels like. 

----
As an aside maybe it would be nice to have a utility that checks a mm config file for best practices (or to even have those documented somewhere which they don't seem to be). Could even potentially be a static webpage where you can upload and get warnings - I may pursue this if i can figure out what the main issues to be avoided are.

pymmcore_plus/core/_mmcore_plus.py
tlambert03: it _won't_ block the native event... but we don't really want to block it unless the native event was triggered specifically by this method.  That's why this pattern is nice

tlambert03: i guess you're worried that we might actually get two signals?

ianhi: Yeah you'd end up with the same signal twice which is probably fine. Although I guess that there is also potential for badness if there is any tight timing. e.g. imagine that there is a callback on a slow hardware switch that results in some fast action, then since we're emitting a signal potentially before the hardware has finished the callback would be triggered prematurely. Although I imagine that `core.watiForSystem` or `waitForDevice` would protect against this.

tlambert03:
I'm going to merge this without the `setConfig` changes.  let's think about that some more

PR #47, opened 2021-08-22
Always call `unloadAllDevices`

ianhi:
This should hopefully protect from having to restart hardware/computers in order to restore connections that we might lose a reference to.

I can add tests but wanted to open first to get feedback if this is an ok strategy - in particular https://stackoverflow.com/a/14986946/835607 has me kinda spooked and I'm not 100% clear that my weakref trickery is an ok thing to do.

Fixes: https://github.com/tlambert03/pymmcore-plus/issues/42

### DIFF
--- a/pymmcore_plus/client/_client.py
+++ b/pymmcore_plus/client/_client.py
@@ -3,7 +3,7 @@ import subprocess
 import sys
 import threading
 import time
-from typing import TYPE_CHECKING, Optional
+from typing import Optional
 
 from loguru import logger
 from Pyro5 import api, core, errors
@@ -11,9 +11,6 @@ from typing_extensions import Protocol
 
 from .. import server
 from .._serialize import register_serializers
-
-if TYPE_CHECKING:
-    from psutil import Process
 
 
 class CallbackProtocol(Protocol):
@@ -45,7 +42,6 @@ class RemoteMMCore(api.Proxy):
         timeout: int = 5,
         verbose: bool = False,
         cleanup_new=True,
-        cleanup_existing=False,
         connected_socket=None,
         callback_class=None,
     ):
@@ -53,9 +49,7 @@ class RemoteMMCore(api.Proxy):
             callback_class = _get_auto_callback_class()
 
         register_serializers()
-        ensure_server_running(
-            host, port, timeout, verbose, cleanup_new, cleanup_existing
-        )
+        ensure_server_running(host, port, timeout, verbose, cleanup_new)
 
         uri = f"PYRO:{server.CORE_NAME}@{host}:{port}"
         super().__init__(uri, connected_socket=connected_socket)
@@ -75,16 +69,6 @@ class RemoteMMCore(api.Proxy):
         if name in ("events",):
             return object.__setattr__(self, name, value)
         return super().__setattr__(name, value)
-
-
-def _get_remote_pid(host, port) -> Optional["Process"]:
-    import psutil
-
-    for proc in psutil.process_iter(["connections"]):
-        for pconn in proc.info["connections"] or []:
-            if pconn.laddr.port == port and pconn.laddr.ip == host:
-                return proc
-    return None
 
 
 def new_server_process(
@@ -111,7 +95,7 @@ def new_server_process(
 
 
 def ensure_server_running(
-    host, port, timeout=5, verbose=False, cleanup_new=True, cleanup_existing=False
+    host, port, timeout=5, verbose=False, cleanup_new=True
 ) -> Optional[subprocess.Popen]:
     """Ensure that a server daemon is running, or start one."""
     uri = f"PYRO:{core.DAEMON_NAME}@{host}:{port}"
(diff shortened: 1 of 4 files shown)


ianhi:
Hmmm, this may not be working for `RemoteMMCore` yet

pymmcore_plus/client/_client.py
```diff
@@ -45,7 +45,7 @@ def __init__(
         timeout: int = 5,
         verbose: bool = False,
         cleanup_new=True,
-        cleanup_existing=True,
+        cleanup_existing=False,
```
ianhi, line 48: Attention here I think this is important to change, but would be curious what the motivations for having it default to true originally were? Having this be true means that if you start mmcore in one process (jupyter) and then connect and close with another (e.g. napari-micromanger) then the connection to the core will be killed. So i'd be inclined to remove the option entirely.

tlambert03: yep i agree, you can default it to `False` and remove the argument.  I think I originally added it because at first I had a lot of stray servers being abandoned, and not being reused.  might have overcompensated :)

ianhi: Removed but also kept the functionality around in `pymmcore_plus.server.try_kill_server` in case it ever comes up in a niche case or when developing.

ianhi:
> I can add tests

I'm not actually sure I can do this. Because doing this requires the process to exit or garbage collection to happen (for CMMCorePlus) and the former would stop the test and that latter is non deterministic when it happens. If you have ideas about how to do this I'd be happy to do them but as it is I'm not totally sure what to do.

ianhi:
Also I think an alternative to adding an `atexit` on the remote process would be to handle the `SIGTERM` signal in CMMCorePlus so that we run `unloadAll` then we'd only have the atexit in one place. Following: https://stackoverflow.com/a/40866947/835607

That said I'm happy with this approach but if that feels more right to you I can change it.

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -41,9 +43,18 @@ def __init__(self, mm_path=None, adapter_paths: ListOrTuple[str] = ()):
         self._canceled = False
         self._paused = False
 
+        # use weakref to avoid atexit keeping us from being
+        # garbage collected
+        self._weak_clean = weakref.WeakMethod(self.unloadAllDevices)
```
tlambert03, line 48: i wonder whether [`weakref.finalize`](https://docs.python.org/3/library/weakref.html#weakref.finalize) might solve this for both the del and the exit case?

tlambert03: otherwise i'm good with this PR.  Can you look into this briefly before we merge?

pymmcore_plus/client/_client.py
```diff
@@ -111,23 +95,22 @@ def new_server_process(
...
-        if cleanup_existing:
-            proc = _get_remote_pid(host, port)
-            if proc is not None:
-                atexit.register(proc.kill)
     except errors.CommunicationError:
         logger.debug("No server found, creating new mmcore server")
         proc = new_server_process(host, port, verbose=verbose)
         if cleanup_new:
             atexit.register(proc.kill)
+            atexit.register(
```
tlambert03, line 111: does order matter here?  should we register `unload` before proc.kill?

pymmcore_plus/core/_mmcore_plus.py
ianhi: > It is important to ensure that func, args and kwargs do not own any references to obj, either directly or indirectly, since otherwise obj will never be garbage collected. In particular, func should not be a bound method of obj.

Seems to imply that we'd still need a `WeakMethod`

pymmcore_plus/client/_client.py
ianhi: Definitely matters - it's important that  `unload` is registered after because `atexit` calls functions in reverse order of registration.

pymmcore_plus/core/_mmcore_plus.py
ianhi: googling for examples perhaps I've misunderstood though

pymmcore_plus/client/_client.py
tlambert03: thanks!

pymmcore_plus/core/_mmcore_plus.py
tlambert03: > Seems to imply that we'd still need a WeakMethod

ah, it sure does.  fine with me to go with what you've got

ianhi: yeah I tried for a bit and couldn't get it working with finalize - seemed like it wasn't always being called for reasons I don't fully get.

tlambert03: thanks for digging! LGTM

PR #48, opened 2021-08-22
Set minimum pymmcore version

ianhi:
This is the version built with proper threading support which is important on many scopes and will crash the callback system.
See https://github.com/micro-manager/pymmcore/issues/49 for more information.


tlambert03 (approved):
thanks! LGTM when tests pass

PR #53, opened 2021-09-13
Add getObjectiveDevices method

ianhi:
For matching state devices that are likely to be an objective.
Needs to match things like the following:

TINosepiece, nosepiece, objective, Objective, ObjectiveTurret etc.

Inspired by: https://github.com/tlambert03/napari-micromanager/pull/43


tlambert03:
👍

tlambert03:
good stuff, thanks!

PR #64, opened 2021-11-30
Fix Aliased signals

ianhi:
Something about the way the signals we're being aliased on the `CMMCoreSignaler` object was not working so the `XYStagePosition` callback was not functioning. It's a bit confusing because I think this used to work but now isn't ??

First commit fails pytest locally, just pushing that to see if it fails CI as well.

Next commit does the aliasing using properties (https://stackoverflow.com/a/11050226/835607) which seems to fix things. I will push that after the tests run.

Fixes: https://github.com/tlambert03/pymmcore-plus/issues/63


ianhi:
And indeed it does fail CI currently: https://github.com/tlambert03/pymmcore-plus/runs/4371884034?check_suite_focus=true

ianhi:
@tlambert03 if this looks ok could I bother you to push a tag so a new release gets made? This makes it a bummer to use napari-micromanager

PR #67, opened 2021-12-03
getOrGuessChannelGroup -> list

fdrgsp:
As for the `guessObjectiveDevices()` method, maybe it is good to have the `getOrGuessChannelGroup()` method returning a list of possible match instead of just the first one that is found (it is possible to have more than one channel group match using the regex `re.compile("(chan{1,2}(el)?|filt(er)?)s?", re.IGNORECASE)`)

The updated method can be like this:
```
chan_group = self.getChannelGroup()
if chan_group:
    return [chan_group]
# not set in core. Try "Channel" and other variations as fallbacks
channel_guess = []
for group in self.getAvailableConfigGroups():
    if self._channel_group_regex.match(group):
        channel_guess.append(group)
return channel_guess
```

### DIFF
--- a/pymmcore_plus/_tests/test_core.py
+++ b/pymmcore_plus/_tests/test_core.py
@@ -331,9 +331,6 @@ def test_get_objectives(core: CMMCorePlus):
 
 def test_guess_channel_group(core: CMMCorePlus):
 
-    chan_group = core.getChannelGroup()
-    assert chan_group == ""
-
     assert core.getOrGuessChannelGroup() == ["Channel"]
 
     with patch.object(core, "getChannelGroup", return_value=""):
--- a/pymmcore_plus/_tests/test_core.py
+++ b/pymmcore_plus/_tests/test_core.py
@@ -340,7 +340,7 @@ def test_guess_channel_group(core: CMMCorePlus):
             core.channelGroup_pattern = 4
 
         # assign a new regex that won't match Channel using a str
-        # this will return all the mm groups, but that's because this a bad regex
+        # this will return all the mm groups except Channel, but that's because this a bad regex
         # to use
         core.channelGroup_pattern = "^((?!(Channel)).)*$"
         assert core.getOrGuessChannelGroup() == [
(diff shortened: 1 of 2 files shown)


pymmcore_plus/_tests/test_core.py
```diff
@@ -330,24 +330,33 @@ def test_get_objectives(core: CMMCorePlus):
...
     with patch.object(core, "getChannelGroup", return_value=""):
-        assert core.getOrGuessChannelGroup() == "Channel"
+        assert core.getOrGuessChannelGroup() == ["Channel"]
 
         with pytest.raises(TypeError):
             core.channelGroup_pattern = 4
 
         # assign a new regex that won't match Channel using a str
-        # this will return Camera, but that's because this a bad regex
+        # this will return all the mm groups, but that's because this a bad regex
```
ianhi, line 343:
```suggestion
        # this will return all the mm groups except Channel, but that's because this a bad regex
```

pymmcore_plus/_tests/test_core.py
```diff
@@ -330,24 +330,30 @@ def test_get_objectives(core: CMMCorePlus):
 
 
 def test_guess_channel_group(core: CMMCorePlus):
-    chan_group = core.getChannelGroup()
-    assert core.getOrGuessChannelGroup() == chan_group
+
+    assert core.getOrGuessChannelGroup() == ["Channel"]
```
ianhi, line 334: Maybe keep the first check that it matches the core method? I think there's some value in that

ianhi (approved):
makes sense to me - one small comment in the test otherwise nice stuff!

pymmcore_plus/_tests/test_core.py
fdrgsp: Sure, I've added it

ianhi:
@fdrgsp if you want then you can combine the commits you've got with this git command:

```
git rebase 7d9df92 -i
```

and then you squash commits by changing the `pick` to be `s` (there should be instructions that pop up)

Doing that so that you end up with two commits:

1. add pixelSizeConfig
2. make guess channel group return a list

can make it easier to review

tlambert03 (approved):
LGTM

PR #71, opened 2022-01-21
Allow more options when loading config

ianhi:
Now will read local files, any file in the _mm_path and absolute paths.


pymmcore_plus/_tests/test_core.py
```diff
@@ -55,6 +56,34 @@ def test_search_paths(core: CMMCorePlus):
         core.setDeviceAdapterSearchPaths("test_path")
 
 
+def test_load_system_config(core: CMMCorePlus):
+    with pytest.raises(FileNotFoundError):
+        core.loadSystemConfiguration("nonexistent")
+
+    test_dir = Path(__file__).parent.absolute()
+    orig_dir = Path.cwd()
+    os.chdir(test_dir)
```
tlambert03, line 65: will this work without using `chdir`, just passing `test_dir /  "local_config.cfg"` to loadSystemConfig?

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -126,15 +126,36 @@ def setDeviceAdapterSearchPaths(self, adapter_paths: ListOrTuple[str]) -> None:
...
+    def loadSystemConfiguration(
+        self, fileName: str | Path = "MMConfig_demo.cfg"
+    ) -> None:
+        """Load a config file.
+
+        Will first search for *fileName* in the following order:
+        1. Relative to the current working directory
+        2. In the micromanager install folder
+        3. As an absolute path
+        """
```
tlambert03, line 138: i feel like relative and absolute paths don't warrant separate treatment... that should be in the path itself (i.e. if it starts with `/` or not)...

and the code itself could be simplified significantly to something like:

```py
fpath = Path(fileName).expanduser()
if not fpath.exists() and not fpath.is_absolute() and self._mm_path:
    fpath = Path(self._mm_path) / fileName
if not fpath.exists():
    raise FileNotFoundError(f"Path does not exist: {fpath}")
return super().loadSystemConfiguration(str(fpath.resolve()))
```

ianhi: I think the issue is that I didn't fully understand what `.resolve()` does. This is much better - will modify

pymmcore_plus/_tests/test_core.py
ianhi: I think it would, except that you can in theory run these tests from any directory - so for that to pass you are assuming that you're running them from the repo top level.

pymmcore_plus/core/_mmcore_plus.py
ianhi: Good mockup - seems like it just works out of the box

pymmcore_plus/_tests/test_core.py
tlambert03: the `test_dir` object is a pathlib.Path object, and line 63 makes it absolutely relative to this file (independent of the current working dir)... so it should work

PR #73, opened 2022-02-09
Start using threads

ianhi:
Working towards: https://github.com/tlambert03/pymmcore-plus/issues/70

This PR does two things

### 1. Add locks
Moving towards more threading introduces opportunities to break things. So I've added locks to the functions that I don't think should be allowed to run concurrently. (I suspect that this was already a possible issue with pyro and maybe we should have always had these)

### 2. Use theads to `run_mda`
So `run_mda` no longer prevents further execution (e.g. freezing UI in napari or running on the fly analysis.).


ianhi:
I also considered doing:

### 3. Make `CMMCorePlus` a singleton

This would make it much easier to do stuff like

```python
# some setup
mmc = CMMCorePlus()

# launch napari-micromanager
```
without having conflicting core object that try to grab hold of the same devices.

In general I think that this would be good, but the option to have multiple is important in case you want to control two micrscopes from the same process or somethign wild like that. I suspect that its also simpler to have two core objects if you have a multi camera setup so we likely shouldn't prevent that.

ianhi:
> I clearly have made something go wrong....

issue was `setXYZ` had a lock on it but also called `setX` which was locked. Solution was to either use an `RLock` or to more simply not lock `setXYZ`

ianhi:
Ok now it's hopefully ready for real. Turns out this does need an `RLock` because otherwise if a callback that calls a locked function is triggered by a locked function it will hang forever. I added a test that should detect cases like that.

ianhi:
A consequence of this threading is that if you launch an MDA from napari you need to use the QCoreCallback object for the signals otherwise things crash. Any thoughts on how you'd like that be structured to avoid duplicating that object between core and client @tlambert03 ?

tlambert03:
sorry, kinda swamped here this week.  Thanks so much for pushing on this though!  I wonder if it might be easier to set up a zoom early next week for you to fill me in on what you're learning?  I'll try to take a look this weekend and have some opinions :)

ianhi:
>  I wonder if it might be easier to set up a zoom early next week for you to fill me in on what you're learning?

@tlambert03 does still seem like a good way to go? If so my calendar is very open

ianhi:
Notes from zoom:

**singletons**
We want a single object per process. Do this with something like:

```python
class MyObject:
    __instance = None

    @classmethod
    def instance(cls):
        if cls.__instance is None:
            cls.__instance = cls()
        return cls.__instance()


MyObject.instance()
```

**wrapt**
Using wrapt is :+1: . Ok to keep class level lock based on only having one instance

**Where do QCallbacks go**
Can put wherever makes sense. Good to avoid duplication between core and client.

**future of client**
we want to keep client around, but may test it less thoroughly, the threaded model is priority going forward


**what to lock**
Don't need to lock `get` methods except maybe for data to prevent race conditions. But should lock any setting of properties.
Could imagine instead having the user lock but can always switch to that.

ianhi:
Before I forget: while I can test this on a real microscope - it's a TE2000, which is old enough that the device adapters don't use threads, so there may be subtle things that break once the device driver gets to spawn it's own threads, and it would be good to be able to test that.

ianhi:
All tests pass locally, disconcerting that they don't here. maybe linux vs non linux?

ianhi:
> All tests pass locally, disconcerting that they don't here. maybe linux vs non linux?

I think this was about version of pyqt. locally I had pyqt installed via conda, which gives 5.12, whereas pip gives 5.15. With pip installing I could replicate. That said I wasn't able to figure out how to deduplicate the QCoreCallback code without running into truly wild errors. So I went with the simple solution of duplicating the signal definitions (it's not much duplication) and now everything works :man_shrugging:. What I was trying is preserved here: https://github.com/ianhi/pymmcore-plus/tree/threading-preserved

pymmcore_plus/_tests/test_core.py
```diff
@@ -19,11 +22,17 @@
...
 
 
 @pytest.fixture
 def core():
     core = CMMCorePlus()
+    # force using psygnal callbacks
+    # necessary until tests can be adapted to work using qtbot
+    core.events = _CMMCoreSignaler()
+    core._callback_relay = pymmcore_plus.core._mmcore_plus.MMCallbackRelay(core.events)
+    core.registerCallback(core._callback_relay)
```
ianhi, line 35: Otherwise I think there end up being some timing issues that require qtbot to resolve.  Ideally we can run tests with both types of callbacks, but until then this is maintaining the status quo of these tests (only using CMMCoreSignaler)

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -43,7 +47,33 @@
...
+    lock = RLock()
+
+    @classmethod
+    def instance(
+        cls, mm_path=None, adapter_paths: ListOrTuple[str] = ()
+    ) -> CMMCorePlus:
+        global _instance
+        if _instance is None:
+            _instance = cls(mm_path, adapter_paths)
+        return _instance
```
tlambert03, line 75:
```suggestion
    __instance = None

    @classmethod
    def instance(
        cls, mm_path=None, adapter_paths: ListOrTuple[str] = ()
    ) -> CMMCorePlus:
        if cls.__instance is None:
            cls.__instance = cls(mm_path, adapter_paths)
        return cls.__instance
```

can we try this as a class attribute rather than a module level global?

ianhi: unfortunately seems like no. See description in commit: https://github.com/tlambert03/pymmcore-plus/pull/73/commits/2bde547f0cf56efc71927d11b05513ec43d853dd

> Has to be this way (single underscore global variable) because
swig seems to prevent us from modifying class attributes, and
double underscore triggers name mangling in a class method:
https://stackoverflow.com/a/59066258/835607

ianhi: concretely:
```python
import pymmcore

class plus(pymmcore.CMMCore):
    __instance = None

    @classmethod
    def instance(cls):
        if cls.__instance is None:
            cls.__instance = cls()
        return cls.__instance
plus.instance()
```
gives

```
Traceback (most recent call last):
  File "/home/ian/Documents/oss/micro/pymmcore-plus/instance.py", line 11, in <module>
    plus.instance()
  File "/home/ian/Documents/oss/micro/pymmcore-plus/instance.py", line 9, in instance
    cls.__instance = cls()
AttributeError: cannot modify read-only attribute 'plus._plus__instance'
```

tlambert03: Ohh indeed. Makes sense

ianhi:
:tada: :tada: hooooorayyy thanks talley!

@tlambert03 can you also cut a release so we can depend on that version in napari-micromanager?

PR #74, opened 2022-02-10
Add a `snap` method and `signal`

ianhi:
1. Add a `snap` method
This is a nice convenience that I've always wanted

2. Add an `imageSnapped` signal

This is will be nice for example when you have opened napari from ipython and are combining scripting and the GUI. e.g. I have some callback that modifies an overlay layer (i.e. cell detection) that I want to run whenever I snap an image either from the gui or from code.


tlambert03 (approved):
I'm ok with this, though I think it's worth being somewhat cautious with adding surface area to the API... (since, for example, it does make one additional thing that imswitch or pycromanager would need to either implement, or we would need to revert eventually).  That said, we've already got a few of those, and this is minor, so LGTM :)

PR #76, opened 2022-02-19
remove caching

ianhi:
in the hopes that it will fix flaky tests

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -36,21 +36,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install -e .[testing]
 
-      - name: Get Date
-        id: get-date
-        run: |
-          echo "::set-output name=date::$(/bin/date -u "+%Y%m")"
-        shell: bash
-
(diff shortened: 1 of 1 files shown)


ianhi:
this craziness: 
image

ianhi:
and that fixed it I guess? weird

tlambert03:
nice!  gonna rerun a couple times

tlambert03:
and again.  this is the vaguely the kind nonsense I was referring to in that [other discussion about caching](https://github.com/tlambert03/napari-micromanager/pull/89#discussion_r796029183).  I have no idea why this shouldn't work, but lets go with it :)

ianhi:
ok now i'm very confused: after merge all but one passed

https://github.com/tlambert03/pymmcore-plus/runs/5255761770?check_suite_focus=true

tlambert03:
darn! :joy:  was this the same error as before?

ianhi:
what's odd is that the thread object should really never be none there. even if the thread had finished I wouldn't have expected that

tlambert03:
yep. totally confused about that too.  very odd indeed

Issue #79, opened 2022-02-20
setPosition lost overloaded signature


tlambert03:
```py
In [1]: from pymmcore_plus import CMMCorePlus

In [2]: mmc = CMMCorePlus()

In [3]: mmc.loadSystemConfiguration()

In [4]: mmc.setPosition(0.0)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [4], in <module>
----> 1 mmc.setPosition(0.0)

File ~/miniconda3/envs/napdev/lib/python3.9/site-packages/wrapt/decorators.py:469, in synchronized.<locals>._synchronized(wrapped, instance, args, kwargs)
    463 @decorator
    464 def _synchronized(wrapped, instance, args, kwargs):
    465     # Execute the wrapped function while the original supplied
    466     # lock is held.
    468     with lock:
--> 469         return wrapped(*args, **kwargs)

TypeError: setPosition() missing 1 required positional argument: 'val'
```

cc @ianhi, can you take a look?

ianhi:
Ah not actually an issue with the decortator but rather an issue with now having `setPosition` be a method on the python side. Previously it was only on C++ side so `setPosition` was overloaded:

![image](https://user-images.githubusercontent.com/10111092/154866245-cb9bc0da-8b02-4d0f-a039-b82cada9f580.png)

but I only implemented one of them (namely: `(str, float)`) in python. Is there a good way to do overloading in python? I found https://docs.python.org/3/library/typing.html#typing.overload but that doesn't work because it doesn't seem to have run time effects.

ianhi:
I don't know if there is a good way to do this in python while maintaining reasonable argument names. I guess we could do *args but hten you lose typing information, or have `stageLabel` sometimes be a float (which isn't great)

ianhi:
I guess that the issues is that `setPosition` would ideally have been `setPosition(position, stageLabel)` which would have allowed for easy translation to python as `setPosition(positoin, stageLabel=None)`, but thats not the owrld we live in :(

tlambert03:
oh good!  that's easier to fix :)

> I guess we could do *args but then you lose typing information

that's where `typing.overload` comes in.  Something like this should work

```python
    @overload
    def setPosition(self, position: float) -> None:
        """Sets the position of the stage in microns."""
    @overload
    def setPosition(self, stageLabel: str, position: float) -> None:
        """Sets the position of the stage in microns."""

    @synchronized(lock)
[...]

PR #82, opened 2022-02-20
restore setPosition overloading

ianhi:
fix: #79

### DIFF
--- a/pymmcore_plus/core/_mmcore_plus.py
+++ b/pymmcore_plus/core/_mmcore_plus.py
@@ -444,15 +444,7 @@ class CMMCorePlus(pymmcore.CMMCore):
     @synchronized(lock)
     def setPosition(self, *args) -> None:
         """Set position of the stage in microns."""
-        n_args = len(args)
-        if n_args == 1:
-            return super().setPosition(args[0])
-        elif n_args == 2:
-            return super().setPosition(args[0], args[1])
-        else:
-            raise ValueError(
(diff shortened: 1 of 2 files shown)


pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -432,9 +433,26 @@ def getZPosition(self) -> float:
...
+        """Set position of the stage in microns."""
+        n_args = len(args)
+        if n_args == 1:
+            return super().setPosition(args[0])
+        elif n_args == 2:
+            return super().setPosition(args[0], args[1])
+        else:
+            raise ValueError(
+                f"setPosition takes exactly one or two arguments. Got {len(args)}"
+            )
```
tlambert03, line 455:
```suggestion
        return super().setPosition(*args)
```

tlambert03: we don't need to raise the value error here.  super() will do it

ianhi: oooh

ianhi:
Thought: we should add a test that the typing of this library is a superset of the pymmcore types. Not quite sure how to implement that, but would prevent similar issues in the future

tlambert03:
Testing typing is rather annoying.  One can use pytest-mypy-plugins to do it (and I do in other libraries to make sure types are working).  But it's slow, and takes a lot of work to keep it working as mypy versions change.  Not worth it for something like this, I should have just checked the super sig

PR #84, opened 2022-02-25
create docs

ianhi:
Would be nice to also enable autobuild on PRs. Though that depends on how you want to host these @tlambert03. If we want to start with readthedocs then I know it's pretty easy.

Also cc @fdrgsp for more eyes on this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,3 +140,5 @@ cython_debug/
 pymmcore_plus/_version.py
 Micro-Manager-*
 .vscode
+
+docs/api
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,11 @@
+version: 2
+python:
+  version: 3.9
+  install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - doc
+
+sphinx:
+  configuration: docs/conf.py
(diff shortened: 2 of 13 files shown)


tlambert03:
thanks @ianhi!  I'm happy to use readthedocs to keep it easy

tlambert03:
this looks amazeballs at first glance.  will pull and build when I have a moment

ianhi:
If you set up readthedocs you won't even need to build! https://docs.readthedocs.io/en/stable/pull-requests.html

tlambert03:
yep, I do like RTD.  use it for a few projects (https://ome-types.readthedocs.io/en/latest/, https://napari-plugin-engine.readthedocs.io/en/latest/)

ianhi:
one todo here is to publicly expose the Signaller object so that it can be captured by autodoc.

ianhi:
> should add docs/api to .gitignore

Ahh the curse of the unpushed commit.

> I'm not a big fan of sphinx_book_theme. would prefer furo, sphinx-material, or pydata-sphinx

I like those as well, with a preference for that order. 

> lets get rid of sphinx-thebe for now. I don't think we need this to interactive, or at the very least, it will need some work to make it so

indeed. removed

tlambert03:
i think this is a great start, so I'm going to merge.  We can continue to iterate in later PRs

tlambert03 (approved):

PR #90, opened 2022-02-26
remove setup.py, add ian as author

tlambert03:


tlambert03:
ok being added to authors @ianhi ?

ianhi:
> ok being added to authors @ianhi ?

Definitely! thanks :)

ianhi (changes requested):
While I'm down with adding me as an author I'm pretty sure think we can't fully remove setup.py until some future advances in setuptools :(. Removing it now would mean that you can't make an editable install. See  https://github.com/pypa/setuptools/issues/2816 for more details.

tlambert03:
Yep, pip added support for editable installs without setup.py towards the end of last year.  I've been removing setup.py from a few projects with no issues so far.

Issue #43, opened 2021-08-14
With RemoteMMCore How to recover if connection to pyro server lost


ianhi:
I've been running into this error recently:

```
ctionResetError                      Traceback (most recent call last)
~\miniconda3\envs\control-pymm\lib\site-packages\Pyro5\socketutil.py in send_data(sock, data)
    181         try:
--> 182             sock.sendall(data)
    183             return

ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host

During handling of the above exception, another exception occurred:

ConnectionClosedError                     Traceback (most recent call last)
~\AppData\Local\Temp/ipykernel_10752/3470287651.py in <module>
----> 1 mmcore.getExposure()

~\miniconda3\envs\control-pymm\lib\site-packages\Pyro5\client.py in __call__(self, *args, **kwargs)
    474         for attempt in range(self.__max_retries + 1):
    475             try:
--> 476                 return self.__send(self.__name, args, kwargs)
    477             except (errors.ConnectionClosedError, errors.TimeoutError):
    478                 # only retry for recoverable network errors

~\miniconda3\envs\control-pymm\lib\site-packages\Pyro5\client.py in _pyroInvoke(self, methodname, vargs, kwargs, flags, objectId)
    217             protocol.log_wiredata(log, "proxy wiredata sending", msg)
    218         try:
--> 219             self._pyroConnection.send(msg.data)
    220             del msg  # invite GC to collect the object, don't wait for out-of-scope
    221             if flags & protocol.FLAGS_ONEWAY:

~\miniconda3\envs\control-pymm\lib\site-packages\Pyro5\socketutil.py in send(self, data)
    430 
    431     def send(self, data: bytes) -> None:
--> 432         send_data(self.sock, data)
    433 
    434     def recv(self, size: int) -> bytes:

~\miniconda3\envs\control-pymm\lib\site-packages\Pyro5\socketutil.py in send_data(sock, data)
    185             raise TimeoutError("sending: timeout")
    186         except socket.error as x:
--> 187             raise ConnectionClosedError("sending: connection lost: " + str(x))
    188     else:
    189         # Socket is in non-blocking mode, use regular send loop.

ConnectionClosedError: sending: connection lost: [WinError 10054] An existing connection was forcibly closed by the remote host
```

[...]

PR #91, opened 2022-02-26
Move remote stuff to subfolder, make pyro5 optional [remote] extra

tlambert03:
the remote functionality brings in a few extra dependencies... so this makes them all optional


ianhi (approved):
Looks good

Only question is: Does it make sense keep it at all? Given the known reliability issues (https://github.com/tlambert03/pymmcore-plus/issues/43) I don't think I would recommend anyone use it. The major uses case I think of is from another computer

tlambert03:
> Does it make sense keep it at all? Given the known reliability issues (https://github.com/tlambert03/pymmcore-plus/issues/43) I don't think I would recommend anyone use it.

While I agree that it shouldn't be the default, I'm not yet ready to drop it entirely.  It _is_ a compelling use case if only we could solve #43 (which i still imagine is possible... I see that issue as me just not using pyro 100% correctly).  So, let's punt on whether we fully remove

ianhi:
Is it possilbe to override `__dir__` as well so that if you are for example `shift-tabbing` in a notebook then Remote will still show up? This works in VSCode but not in jupyterlab, or any other jedi based systems I imagine.

ianhi:
Doing this brings back autocomplete for jedi:

```python
dir_ = dir() + ["RemoteMMCore", "server"]
def __dir__():
    return dir_
```

but it also feels somewhat wrong.

tlambert03:
> Doing this brings back autocomplete for jedi:
> but it also feels somewhat wrong.

let's go with

```py
def __dir__():
    return __all__
```

ianhi:
>      let's go with

> def __dir__():
>    return __all__

I tried that as well but then you lose all the normal dunder methods on shift tab :(.

tlambert03:
ah, ok, then:
```python
def __dir__():
    return list(globals()) + ['RemoteMMCore', 'server']
```

ianhi:
Just checked - that works!

PR #93, opened 2022-02-28
fix python version in RTD build

tlambert03:
looks like RTD doesn't support 3.9 yet


tlambert03:
note, I also had to change `.readthedocs.yml` to `.readthedocs.yaml`, and add a couple more deps (`sphinx-panels`, and `sphinx-togglebutton`) to the extras_require.
this was done directly on main

PR #94, opened 2022-02-28
Doc follow ups

ianhi:
We should also:

1. Add a link to the github about section
2. Have RTD build on PRs (only @tlambert03 can enable that)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # pymmcore-plus
 
+[![Documentation Status](https://readthedocs.org/projects/pymmcore-plus/badge/?version=latest)](https://pymmcore-plus.readthedocs.io/en/latest/?badge=latest)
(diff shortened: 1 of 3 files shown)


tlambert03:
it's already up on rtd https://pymmcore-plus.readthedocs.io/en/latest/

tlambert03:
enabled

ianhi:
hmm linkcode works locally but not on readthedocs :/

tlambert03:
> hmm linkcode works locally but not on readthedocs :/

not following... what is link code?

ianhi:
> not following... what is link code?

It's the function in conf.py that makes the [Source] in the api docs:
image

link to github.


also I should say that it's fixed now :)

ianhi:
Previously it was rendering as these broken links: https://github.com/tlambert03/pymmcore-plus/blob/envs/latest/lib/python3.8/site-packages/pymmcore_plus/core/_constants.py#L60-L67

PR #97, opened 2022-03-02
Ensure propertyChanged event occurs for setState and setStateLabel

tlambert03:
Akin to #38, when `core.setState()` is called it will _sometimes_ cause a `core.events.propertyChanged` event to be emitted, in a device-dependent manor.  (for example, in the demo config, it works for `Objective`, but not for `Dichroic`).  This pulls the logic from #38 into a more generally useful context manager and uses it for `setState` and `setStateLabel` as well

### DIFF
--- a/pymmcore_plus/_tests/test_core.py
+++ b/pymmcore_plus/_tests/test_core.py
@@ -353,6 +353,28 @@ def test_set_property_events(core: CMMCorePlus):
     mock.assert_called_once_with("Camera", "AllowMultiROI", "1")
 
 
+def test_set_state_events(core: CMMCorePlus):
+    mock = MagicMock()
+    core.events.propertyChanged.connect(mock)
+    assert core.getState("Objective") == 1
+    core.setState("Objective", 3)
+    mock.assert_has_calls(
+        [
+            call("Objective", "State", "3"),
+            call("Objective", "Label", "Nikon 20X Plan Fluor ELWD"),
+        ]
+    )
+    assert core.getState("Objective") == 3
+
+    mock.reset_mock()
+    assert core.getState("Dichroic") == 0
+    core.setState("Dichroic", 1)
+    mock.assert_has_calls(
+        [call("Dichroic", "State", "1"), call("Dichroic", "Label", "Q505LP")]
+    )
+    assert core.getState("Dichroic") == 1
+
+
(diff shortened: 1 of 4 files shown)


pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -111,38 +112,25 @@ def __del__(self):
...
-        ):  # block the native event.
-            super().setProperty(label, propName, propValue)
-        after = super().getProperty(label, propName)
-        if before != after:
-            self.events.propertyChanged.emit(label, propName, after)
+    @synchronized(lock)
+    def setStateLabel(self, stateDeviceLabel: str, stateLabel: str) -> None:
+        """Set state (by label) on stateDeviceLabel, with reliable event emission."""
+        with self._property_change_emission_ensured(
+            stateDeviceLabel, ("State", "Label")
```
ianhi, line 131: What happens if a state doesn't have an associated label? I suspect that a blank string will be sent, but we should check that this doesn't cause an error before merging.

pymmcore_plus/_tests/test_core.py
```diff
@@ -353,6 +353,28 @@ def test_set_property_events(core: CMMCorePlus):
...
 
+def test_set_state_events(core: CMMCorePlus):
+    mock = MagicMock()
+    core.events.propertyChanged.connect(mock)
+    assert core.getState("Objective") == 1
+    core.setState("Objective", 3)
+    mock.assert_has_calls(
+        [
+            call("Objective", "State", "3"),
+            call("Objective", "Label", "Nikon 20X Plan Fluor ELWD"),
```
ianhi, line 364: Mentioning to expect both of these and only handle one of them would be great to include in a `working with properties and state devices` example.

ianhi (commented):
One concern about what happens for a state device with no labels - which I can check. Otherwise Looks good!

tlambert03:
Good points all around!  Thanks, will address tomorrow.

pymmcore_plus/core/_mmcore_plus.py
ianhi: Looks like they they default to `State-1` and all is well

ianhi (approved):
No longer concerns about unlabelled states. So this is good to go from my perspective.

Suggestions about docstring + example stand though.

PR #102, opened 2022-03-04
Switch to user swappable mda-engines

ianhi:
As discussed in: https://github.com/tlambert03/pymmcore-plus/issues/101
Not polished yet - Opening for feedback from @tlambert03 

The big thing I struggled with was the relationship between core and the mda object should be. The mda engine needs to have access to core methods in order to function but then you can end up with this awful thing: `core._mda_engine._mmc._mda_engine._mmc.....`

If we decide to got this way then this should also get:
1. update the client to reflect the new structure.
2. Start of an `mda_tests` file
3. add an example showing how to define your own MDA engine.

### DIFF
--- a/pymmcore_plus/_tests/conftest.py
+++ b/pymmcore_plus/_tests/conftest.py
@@ -1,22 +1,20 @@
 import pytest
 
 import pymmcore_plus
-
 from pymmcore_plus.core.events import CMMCoreSignaler, QCoreSignaler
 from pymmcore_plus.mda.events import MDASignaler, QMDASignaler
+
 
 @pytest.fixture(params=["QSignal", "psygnal"], scope="function")
 def core(request):
     core = pymmcore_plus.CMMCorePlus()
     if request.param == "psygnal":
-        core.events = pymmcore_plus.CMMCoreSignaler()
+        core.events = CMMCoreSignaler()
         core.mda._events = MDASignaler()
     else:
         core.events = QCoreSignaler()
         core.mda._events = QMDASignaler()
-    core._callback_relay = pymmcore_plus.core._mmcore_plus.MMCallbackRelay(
-        core.events
-    )
+    core._callback_relay = pymmcore_plus.core._mmcore_plus.MMCallbackRelay(core.events)
     core.registerCallback(core._callback_relay)
     if not core.getDeviceAdapterSearchPaths():
         pytest.fail(
--- a/pymmcore_plus/_tests/conftest.py
+++ b/pymmcore_plus/_tests/conftest.py
@@ -2,16 +2,22 @@ import pytest
 
 import pymmcore_plus
 
+from pymmcore_plus.core.events import CMMCoreSignaler, QCoreSignaler
+from pymmcore_plus.mda.events import MDASignaler, QMDASignaler
 
 @pytest.fixture(params=["QSignal", "psygnal"], scope="function")
 def core(request):
     core = pymmcore_plus.CMMCorePlus()
     if request.param == "psygnal":
         core.events = pymmcore_plus.CMMCoreSignaler()
-        core._callback_relay = pymmcore_plus.core._mmcore_plus.MMCallbackRelay(
-            core.events
-        )
-        core.registerCallback(core._callback_relay)
+        core.mda._events = MDASignaler()
+    else:
+        core.events = QCoreSignaler()
+        core.mda._events = QMDASignaler()
+    core._callback_relay = pymmcore_plus.core._mmcore_plus.MMCallbackRelay(
+        core.events
+    )
+    core.registerCallback(core._callback_relay)
     if not core.getDeviceAdapterSearchPaths():
         pytest.fail(
             "To run tests, please install MM with `python -m pymmcore_plus.install`"
--- a/pymmcore_plus/core/_mmcore_plus.py
+++ b/pymmcore_plus/core/_mmcore_plus.py
@@ -469,6 +469,10 @@ class CMMCorePlus(pymmcore.CMMCore):
         Thread
             The thread the MDA is running on.
         """
+        if self._mda_engine.is_running():
+            raise ValueError(
+                "Cannot start an MDA while the previous MDA is still running."
+            )
         th = Thread(target=self._mda_engine.run, args=(sequence,))
         th.start()
         return th
--- a/pymmcore_plus/core/_mmcore_plus.py
+++ b/pymmcore_plus/core/_mmcore_plus.py
@@ -93,7 +93,7 @@ class CMMCorePlus(pymmcore.CMMCore):
         self._callback_relay = MMCallbackRelay(self.events)
         self.registerCallback(self._callback_relay)
 
-        self._mda_engine = MDAEngine(self)
+        self._mda_engine = MDAEngine()
 
         self._objective_regex = _OBJECTIVE_DEVICE_RE
         self._channel_group_regex = _CHANNEL_REGEX
--- a/pymmcore_plus/core/_mmcore_plus.py
+++ b/pymmcore_plus/core/_mmcore_plus.py
@@ -476,7 +476,8 @@ class CMMCorePlus(pymmcore.CMMCore):
     def register_mda_engine(self, engine):
         """
         Set the MDA Engine to be used on ``run_mda``. This will unregister
-        the previous engine and emit an ``mdaEngineRegistered`` signal.
+        the previous engine and emit an ``mdaEngineRegistered`` signal. The
+        current Engine must not be running an MDA in order to register a new engine.
 
         Parameters
         ----------
@@ -485,6 +486,12 @@ class CMMCorePlus(pymmcore.CMMCore):
         """
         if not isinstance(engine, PMDAEngine):
             raise TypeError("Engine does not conform to the Engine protocol.")
+        if self._mda_engine.is_running():
+            raise ValueError(
+                "Cannot register a new engine when the current engine is running "
+                "an acquistion. Please cancel the current engine's acquistion "
+                "before registering"
+            )
         previous_engine, self._mda_engine = self._mda_engine, engine
         self.events.mdaEngineRegistered.emit(engine, previous_engine)
 
--- a/pymmcore_plus/core/_mmcore_plus.py
+++ b/pymmcore_plus/core/_mmcore_plus.py
@@ -93,7 +93,7 @@ class CMMCorePlus(pymmcore.CMMCore):
         self._callback_relay = MMCallbackRelay(self.events)
         self.registerCallback(self._callback_relay)
 
-        self._mda_engine = MDAEngine()
+        self._mda_engine = MDAEngine(self)
 
         self._objective_regex = _OBJECTIVE_DEVICE_RE
         self._channel_group_regex = _CHANNEL_REGEX
--- a/pymmcore_plus/core/_mmcore_plus.py
+++ b/pymmcore_plus/core/_mmcore_plus.py
@@ -458,9 +458,8 @@ class CMMCorePlus(pymmcore.CMMCore):
 
     def register_mda_engine(self, engine):
         if not isinstance(engine, PMDAEngine):
-            raise TypeError("engine must conform to the Engine protocol.")
-        previous_engine = self._mda_engine
-        self._mda_engine = engine
+            raise TypeError("Engine does not conform to the Engine protocol.")
+        previous_engine, self._mda_engine = self._mda_engine, engine
         self.events.mdaEngineRegistered.emit(engine, previous_engine)
 
     def _fix_image(self, img: np.ndarray) -> np.ndarray:
--- a/pymmcore_plus/core/_mmcore_plus.py
+++ b/pymmcore_plus/core/_mmcore_plus.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import atexit
 import os
 import re
-import time
 import weakref
 from contextlib import contextmanager
 from datetime import datetime
@@ -30,10 +29,11 @@ from psygnal import SignalInstance
 from wrapt import synchronized
 
 from .._util import find_micromanager
+from ..mda import MDAEngine, PMDAEngine
 from ._config import Configuration
 from ._constants import DeviceDetectionStatus, DeviceType, PropertyType
 from ._metadata import Metadata
-from .events import CMMCoreSignaler, _get_auto_callback_class
+from .events import CMMCoreSignaler, _get_auto_core_callback_class
 
 if TYPE_CHECKING:
     import numpy as np
@@ -89,11 +89,11 @@ class CMMCorePlus(pymmcore.CMMCore):
         if adapter_paths:
             self.setDeviceAdapterSearchPaths(adapter_paths)
 
-        self.events = _get_auto_callback_class()()
+        self.events = _get_auto_core_callback_class()()
         self._callback_relay = MMCallbackRelay(self.events)
         self.registerCallback(self._callback_relay)
-        self._canceled = False
-        self._paused = False
+
+        self._mda_engine = MDAEngine(self)
 
         self._objective_regex = _OBJECTIVE_DEVICE_RE
         self._channel_group_regex = _CHANNEL_REGEX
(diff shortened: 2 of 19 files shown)


ianhi:
From a user perspective the main (only?) change will be that you now type `core.mda` rather then `core.` when accessing mda related things.

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -447,81 +447,21 @@ def getCameraChannelNames(self) -> Tuple[str, ...]:
...
-            self.events.frameReady.emit(img, event)
-
-        logger.info("MDA Finished: {}", sequence)
-        self.events.sequenceFinished.emit(sequence)
+    def register_mda_engine(self, engine):
+        if not isinstance(engine, PMDAEngine):
+            raise TypeError("engine must conform to the Engine protocol.")
+        previous_engine = self._mda_engine
+        self._mda_engine = engine
+        self.events.mdaEngineRegistered.emit(engine, previous_engine)
```
ianhi, line 464: added so that downstream things like napari-micromanager can unregister their callbacks with the old engine and attach to the new one.

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -447,81 +447,21 @@ def getCameraChannelNames(self) -> Tuple[str, ...]:
...
-            self.snapImage()
-            img = self.getImage()
-
-            self.events.frameReady.emit(img, event)
-
-        logger.info("MDA Finished: {}", sequence)
-        self.events.sequenceFinished.emit(sequence)
+    def register_mda_engine(self, engine):
+        if not isinstance(engine, PMDAEngine):
+            raise TypeError("engine must conform to the Engine protocol.")
```
ianhi, line 461: should have a test that this fails as expected. Also better error text

ianhi:
image
nice :/  they all passed locally I swear!

ianhi:
ah this is actually a bit complicated to add to remote (or at least I'm not sure exactly how to do) So going to put that off until we solidify how the local variant should work.

pymmcore_plus/mda/_engine.py
```diff
@@ -0,0 +1,113 @@
...
+from useq import MDASequence
+
+from .events import PMDASignaler, _get_auto_MDA_callback_class
+
+if TYPE_CHECKING:
+    from ..core import CMMCorePlus
+
+
+@runtime_checkable
+class PMDAEngine(Protocol):
```
tlambert03, line 14: `Protocol` also actually works as an ABC.  So, if we add some `@abstractmethod` decorators here, someone could (optionally) use it as a base class and make sure their protocol is in spec.  It still supports structural sub-typing though, so they don't have to use.

So this could look like 

```python

@runtime_checkable
class PMDAEngine(Protocol):
    @abstractmethod
    def __init__(self, mmc: "CMMCorePlus") -> None:
        ...

    @abstractproperty
    def events(self) -> PMDASignaler:
        """Add docstring prescribing behavior."""

    @abstractmethod
    def cancel(self):
        """Add docstring prescribing behavior."""

    @abstractmethod
    def toggle_pause(self):
        """Add docstring prescribing behavior."""

    @abstractmethod
    def run(self, sequence: MDASequence):
        """Add docstring prescribing behavior."""
```

and then lower (optional, but nice to immediately show that it implements the protocol)

```python
class MDAEngine(PMDAEngine):
```

pymmcore_plus/mda/_engine.py
```diff
@@ -0,0 +1,113 @@
...
+        ...
+
+    def run(self, sequence: MDASequence):
+        ...
+
+
+class MDAEngine:
+    def __init__(self, mmc: "CMMCorePlus") -> None:
+        self.events = _get_auto_MDA_callback_class()()
+        # we can't use the Instance approach here
```
tlambert03, line 30: if we really wanted to, we could set self._mmc to `None`  and then have a property that gets the global instance

```
@property
def mmc(self):
    return self._mmc or CMMCorePlus.instance()
```

tlambert03 (commented):
really like the direction this is going!

> The big thing I struggled with was the relationship between core and the mda object should be. The mda engine needs to have access to core methods in order to function but then you can end up with this awful thing: core._mda_engine._mmc._mda_engine._mmc.....

I think it's close to fine as it is.  This part is really an internal implementation detail right? (since the whole point is to allow for an mda object that uses some other core/engine?).  I did leave a small note about how you could make `mmc` optional in the constructor of `MDAEngine`, and then retrieve the global singleton afterwards with a property (only when needed)

Also not worried about the recursive pointers.

let me know when you feel like this is settling into finished, or if you are mostly just at the point where you're fighting with CI

pymmcore_plus/mda/_engine.py
ianhi: The other way I think is to call `instance` in the `run` method which is the only place it actually gets used.

pymmcore_plus/mda/_engine.py
ianhi: oh i re-read what you wrote. I skimmed too quickly

pymmcore_plus/mda/_engine.py
tlambert03: yep, that works too!

pymmcore_plus/mda/_engine.py
tlambert03: > I think it might also be nice to include a method like the waiting loop as a method that can be used by a subclass. Is it ok to have a method with implementation on the ABC or should that really go somewhere else?

yeah, it's technically ok... but if you're imagining something like a hook that gets called somewhere inside of `run`, then i think it actually makes the "protocol" part a lot more complicated.  Since whether that hook gets called or not really depends on whether the `run` method calls it: making it more of an implementation detail of the engine, rather than an enforceable part of the protocol

pymmcore_plus/mda/_engine.py
ianhi: or a third option of the `core.mda` being lazily created.

tlambert03: I think I like keeping this in mda.  I like the optional input in the `__init__`... whether you use the `@property` for `mmc`, or just use `self._mmc or CMMCorePlus.instance()` in the run method, I don't mind

ianhi:
I'm really struggling with getting this to work through pyro. I can't seem to get past this:

```
AttributeError: remote object 'PYRO:pymmcore_plus.CMMCorePlus@127.0.0.1:54333' has no exposed attribute or method 'mda'
```

I think because the `wrap_for_pyro` isn't recognizing properties. But even if I go in an add some logic to make sure it's on the proxy class the RemoteMMCore doesn't seem to pick it up.

ianhi:
Latest commit is my current best effort at getting remote to work.

pymmcore_plus/remote/_util.py
```diff
@@ -28,13 +28,32 @@ def _f(self, *args, **kwargs):
...
                     setattr(_dict_[k], attr, getattr(v, attr))
+        elif isinstance(v, property):
+            if "mda" in k:
+                print("here")
+            _dict_[k] = _proxy_property(k)
 
     _dict_["__init__"] = __init__
-    return type(f"{cls.__name__}Proxy", (), _dict_)
+    proxy = type(f"{cls.__name__}Proxy", (), _dict_)
+    print("mda" in _dict_)
```
ianhi, line 57: This returns true, which is why I'm confused about it not showing up on RemoteMMCore

tlambert03:
> Any advice on getting that to work?

I say ditch it for now... skip the remote tests and make any assertions you need/want in the code to ensure that we're running locally

tlambert03:
> AttributeError: remote object 'PYRO:pymmcore_plus.CMMCorePlus@127.0.0.1:54333' has no exposed attribute or method 'mda'

yeah, wrap_for_pyro was really quite a challenge with SWIG-compiled classes.  I asked a question about that https://github.com/irmen/Pyro5/issues/41 ... but I get the sense that the library made the assumption that it's working with pure python classes that behave normally.  A reasonable assumption, but definitely a big difficulty in our case.  

fine with me not to support that here

ianhi:
> fine with me not to support that here

You mean that `Remote` cannot use mda? because I think that's more or less what that would amount - because you can't access the mda events.

tlambert03:
>You mean that Remote cannot use mda?

yep... pretty much.  I really do think that until #43 is fixed, the whole thing is more or less useless anyway.  If some great need returns for it in the future, then I would 1) re-evaluate whether there is a different IPC strategy that might be better 2) fix #43 then 3) come back to fixing mda

ianhi:
> yep... pretty much. I really do think that until https://github.com/tlambert03/pymmcore-plus/issues/43 is fixed, the whole thing is more or less useless anyway. If some great need returns for it in the future, then I would 1) re-evaluate whether there is a different IPC strategy that might be better 2) fix https://github.com/tlambert03/pymmcore-plus/issues/43 then 3) come back to fixing mda

Ok I'll drop those commits - and we can add a note in the docs about this limitation.

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -447,81 +447,20 @@ def getCameraChannelNames(self) -> Tuple[str, ...]:
...
-            # acquire
-            self.waitForSystem()
-            self.snapImage()
-            img = self.getImage()
-
-            self.events.frameReady.emit(img, event)
-
-        logger.info("MDA Finished: {}", sequence)
-        self.events.sequenceFinished.emit(sequence)
+    def register_mda_engine(self, engine):
```
tlambert03, line 459: docstring

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -447,81 +447,20 @@ def getCameraChannelNames(self) -> Tuple[str, ...]:
     def snapImage(self) -> None:
         return super().snapImage()
 
+    @property
+    def mda(self):
+        return self._mda_engine
+
     def run_mda(self, sequence: MDASequence) -> Thread:
```
tlambert03, line 454: i know it previously didn't have one, but could you add a very brief docstring here?

pymmcore_plus/mda/_engine.py
```diff
@@ -0,0 +1,138 @@
...
+    @property
+    @abstractmethod
+    def events(self) -> PMDASignaler:
+        """Return the MDA events object."""
+        ...
+
+    @abstractmethod
+    def cancel(self):
+        """Cancel the MDA."""
+        ...
```
tlambert03, line 25:
```suggestion
```

pymmcore_plus/mda/_engine.py
```diff
@@ -0,0 +1,138 @@
...
+    from ..core import CMMCorePlus
+
+
+@runtime_checkable
+class PMDAEngine(Protocol):
+    @property
+    @abstractmethod
+    def events(self) -> PMDASignaler:
+        """Return the MDA events object."""
+        ...
```
tlambert03, line 20:
```suggestion
```
(can remove `...` when you have a docstring.)

pymmcore_plus/mda/_engine.py
```diff
@@ -0,0 +1,138 @@
...
+
+    @abstractmethod
+    def cancel(self):
+        """Cancel the MDA."""
+        ...
+
+    @abstractmethod
+    def toggle_pause(self):
+        """Switch whether the MDA is paused."""
+        ...
```
tlambert03, line 30:
```suggestion
```

pymmcore_plus/mda/_engine.py
```diff
@@ -0,0 +1,138 @@
...
+
+    @abstractmethod
+    def is_paused(self) -> bool:
+        """Returns whether the acquistion is currently paused."""
+        ...
```
tlambert03, line 35:
```suggestion
```

pymmcore_plus/mda/_engine.py
```diff
@@ -0,0 +1,138 @@
...
+
+    @abstractmethod
+    def run(self, sequence: MDASequence):
+        """Start the acquisition loop blocking the current thread."""
+        ...
```
tlambert03, line 40:
```suggestion
```

tlambert03 (approved):
i know it's still draft, but it looks like tests are passing and the code is looking great!  Left a few minor comments, but  approving now.  Feel free to merge when ready (or ping me if you want more input)

pymmcore_plus/core/_mmcore_plus.py
ianhi: good call

pymmcore_plus/mda/_engine.py
```diff
@@ -0,0 +1,143 @@
...
+        self._canceled = False
+        self._paused = False
+        self._running = False
+
+    @property
+    def events(self) -> PMDASignaler:
+        return self._events
+
+    def is_running(self) -> bool:
+        return self._running
```
ianhi, line 55: only real heads up is that I added this so that we can make sure an MDA isn't running when the user registers a new engine. otherwise you can end up losing track of it and have no way to cancel it.

pymmcore_plus/_tests/conftest.py
```diff
@@ -1,17 +1,21 @@
...
-        core._callback_relay = pymmcore_plus.core._mmcore_plus.MMCallbackRelay(
-            core.events
-        )
-        core.registerCallback(core._callback_relay)
+        core.events = CMMCoreSignaler()
+        core.mda._events = MDASignaler()
+    else:
+        core.events = QCoreSignaler()
+        core.mda._events = QMDASignaler()
+    core._callback_relay = pymmcore_plus.core._mmcore_plus.MMCallbackRelay(core.events)
```
ianhi, line 17: also of note - I don't think we were garunteed to get the QSignaler before. At least locally I wasn't getting it.

ianhi:
> and we can add a note in the docs about this limitation.

Remote is actually never mentioned in the docs - so don't think it makes sense to add one for this.

> i know it's still draft, but it looks like tests are passing and the code is looking great! 
> Feel free to merge when ready

:grinning: will do in a bit

tlambert03:
lovely!

ianhi:
hooray! Thanks for the help talley!


Also heads up to @henrypinkard - you can make an object that conforms to the Engine protocol and then pass the actual acquisition through to micromanager

also may be useful for @ptbrown1729 to implement your own engine to use your DAQ devices. Feel free to ping me for help with that - I may end up doing that for my setup next week

henrypinkard:
Paused, cancel, etc make sense but what is this thing:

class PMDASignaler(*args, **kwargs)

ianhi:
It's anything that has the right names of the mda signals. e.g. this psygnal version:
https://github.com/tlambert03/pymmcore-plus/blob/main/pymmcore_plus/mda/events/_psygnal.py

which you can just import `from pymmcore_plus.mda.events import MDASignaler` and use

henrypinkard:
It seems to me that this would be a confusing addition to the pycromanager api. Is it possible to implement a translation class within this project that translates this api to the pycromanager API?

tlambert03:
The `PMDASignaler` is a `typing.Protocol` that declares the event/updating interface that an engine needs to provide.  We need some way for pycromanager to tell us what's happening... and that tells you the names of the signals to emit.  This goes back to the psygnal event emitter thing...

tlambert03:
(you don't actually need to subclass that or anything, just need to provide the interface)

henrypinkard:
I see. But this doesn't quite answer my question.

Pycromanager doesn't use events anywhere. The equivalent things happen with callback functions (when they happen at all). I am hesitant to add this to pycromanager, because it duplicates functionality with a different, and nonessential mechanism, which I think would make the API more confusing to user using pycromanager alone.

What I am suggesting is that a compatibility layer that translates the the mechanisms used by pycromanger to this event-based system live in pymmcore-plus, since this seems to be where the specification of an event-based acquisition engine lives. Does that make sense?

tlambert03:
It does,  maybe @ianhi has an idea here, or perhaps we can zoom about this.  I understand that pycromanager doesn't use events anywhere yet, largely because all of the event-related stuff that would signal to the GUI is on the Java side where the gui was written.  I can't immediately think of how we would know that pycromanager has "done anything" without it notifying us somehow (i.e. using an Event).

It would be very much akin to what you did [here](https://github.com/micro-manager/pycro-manager/blob/10057dd5efc7167a3389383e8812e69b1513c1db/pycromanager/napari_util.py#L39-L44)

```python
                if not hasattr(napari_signaller, 'old_shape') or \
                        napari_signaller.old_shape.size != shape.size or \
                        np.any(napari_signaller.old_shape != shape):
                    image = dataset.as_array()
                    napari_signaller.old_shape = shape
                    yield image
```

except that rather than yield image, you would emit a signal `frameReady.emit(image)`.  This doesn't have to mean you're extending the public facing pycro-manager API... it's just a good way for the programs to meet half-way with loose coupling

henrypinkard:
> largely because all of the event-related stuff that would signal to the GUI is on the Java side where the gui was written. I can't immediately think of how we would know that pycromanager has "done anything" without it notifying us somehow (i.e. using an Event).

Everything is using callbacks (on python and java sides), which do essentially the same thing. I try to studiously avoid event-based systems because I find them to be much harder to debug at compile time

> This doesn't have to mean you're extending the public facing pycro-manager API... it's just a good way for the programs to meet half-way with loose coupling

That makes sense.


Yeah maybe lets set up a zoom next week? I feel like we might be able to knock this all out in a couple hours

tlambert03:
Call them callbacks, call them events, it's the same thing here :)

You connect the callback to the signal emitter.

ianhi:
> What I am suggesting is that a compatibility layer that translates the the mechanisms used by pycromanger to this event-based system live in pymmcore-plus, since this seems to be where the specification of an event-based acquisition engine lives. Does that make sense?

So was I thinking this would make the most sense to have either in a new package: `pycromanager-plus-bridge` or as a sub package (like dataset) of pycromanager. So you could do:

```python
from pycromanager import Core
from pycromanager.pymmcore_plus_bridge import EngineWrapper

core = Core()
engine = EngineWrapper(core)
CMMCorePlus.instance().register_mda_engine(engine)
```
and then it would be used by napari-micromanager etc. The engine wrapper would do the small work of translating pycromanager callbacks to the required signals.


As for why I think that that should live on pycromanager's side

> Yeah maybe lets set up a zoom next week? I feel like we might be able to knock this all out in a couple hours

I can zoom - I think you have my email right? We can coordinate there

ianhi:
This new pycromanager subpackage (or freestanding package) I'm proposing could also provide a mixin class that would benefit from all the pymmcore-plus additions while still calling out to java for everything provided there. Basically this:

```python
class pycroCore:
    def somemethod(self):
        print("Calling pycromanager somemethod")


class CMMCorePlus:
    def somemethod(self):
        print("CMMCore somemethod")

    def cmmonly(self, somearg):
        """A method only on CMMCorePlus"""
        self.somemethod()
        print(f"got {somearg}")


class pycroCoreBridge(pycroCore, CMMCorePlus):
    ...


core = pycroCoreBridge()
core.somemethod()
core.cmmonly("blarg")
```

then all that would needed is some workin into pycroCore init to convert pycromanager style callbacks to CMMCorePlus style and I'm pretty sure it would just work every where that CMMCorePlus does.


Also now that I think about it more. I think a standalone package makes the most sense. It reflects the reality that it's a bridge between libraries that have more of sibling than parent child relationship. Once you put the bridge in one or the other then there's an awkward coupling that's taken place.

(I would give a more complete and tested example - but unfortunately I've never been able to use pycromanager on my laptop because I can't install full micro-manager)

PR #110, opened 2022-03-07
reconfigure MDAEngine to make it easier to subclass

ianhi:
This breaks up the MDAEngine `run` loop into separate methods which can be reused by subclasses. That way if all they want to do is extend some hardware stuff (like I do :) ) then they can still rely on our implementation of waiting for the next event.

The only behavior change is that the `self._running` is now set when the sequence is finished rather than immediately on cancellation. I think this behavior is more correct, but did require changing what we wait for in some of the tests.

### DIFF
--- a/pymmcore_plus/mda/_engine.py
+++ b/pymmcore_plus/mda/_engine.py
@@ -84,10 +84,12 @@ class MDAEngine(PMDAEngine):
         Toggle the paused state of the current acquisition.
 
         To get whether the acquisition is currently paused use the
-        ``is_paused`` method.
+        ``is_paused`` method. This method is a no-op if no acquistion is
+        currently underway.
         """
-        self._paused = not self._paused
-        self._events.sequencePauseToggled.emit(self._paused)
+        if self._running:
+            self._paused = not self._paused
+            self._events.sequencePauseToggled.emit(self._paused)
 
     def is_paused(self) -> bool:
         """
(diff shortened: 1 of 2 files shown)


pymmcore_plus/mda/_engine.py
```diff
@@ -111,34 +150,69 @@ def check_canceled():
...
+        Parameters
+        ----------
+        sequence : MDASequence
+            The sequence that was finished.
+        """
+        logger.info("MDA Finished: {}", sequence)
+        self._running = False
+        self._events.sequenceFinished.emit(sequence)
+
+    def run(self, sequence: MDASequence) -> None:
```
tlambert03, line 201: i know we didn't have one before, but since you added such nice docstrings elsewhere, here too? :)

tlambert03 (approved):
very nice!

pymmcore_plus/mda/_engine.py
ianhi: good point. I even went ahead and added them to all the method

ianhi:
ugh looks like the thread that is somehow mysteriously None bug is back:

```
     
        with qtbot.waitSignal(core.mda._events.sequenceFinished):
>           core.run_mda(mda)
E           AttributeError: 'NoneType' object has no attribute 'start'
```

I'm at loss of what to do here. mark it as a flaky for pytest?

ianhi:
It also seems to pass every single time locally

tlambert03:
yep!  will merge now.  but for future ref: if I approve something, feel free to merge when ready

PR #111, opened 2022-03-07
add devicePropertyChanged and valueChanged signal on `DeviceProperty` object.

tlambert03:
adds a convenient way to connect a callback to a specific device property (or all properties on a specific device):

```python
core.events.devicePropertyChanged("Camera", "Gain").connect(callback)
core.events.devicePropertyChanged("Camera").connect(callback)  # any camera property
```

also added to `DeviceProperty` objects:
```python
prop = DeviceProperty('Objective', 'Label', core)
prop.valueChanged.connect(callback)
```

also added to `Device` objects:
```python
device = Device('Objective', core)
device.propertyChanged.connect(callback)
# or
device.propertyChanged("Gain").connect(callback)
```


tlambert03:
@ianhi, better suggestion for `connect_change_callback` name?

ianhi:
hmm, is there a technical reason to not create a Signal on these objects? Then we could call it something like valueChanged, and more importantly it would be consistent with the way you attach callbacks in the rest of the package. 

But for this approach to things this name seems pretty ok to me.

tlambert03:
> hmm, is there a technical reason to not create a Signal on these objects? Then we could call it something like valueChanged, and more importantly it would be consistent with the way you attach callbacks in the rest of the package.

the main reason I avoided this is that I don't want callbacks to be spread over many places.  It's the core that's really in charge of the model, and `core.events.propertyChanged` that we really care about.  If we added a new `Signal()` to this class, we'd need to connect it up to core anyway, to make sure it got emitted when `core.events.propertChanged` got emitted, and vice versa.  We could conceivably have some "emitter like" object on `DeviceProperty` just for the syntax `property.changed.connect(some_callback)` ... but what we _really_ need here is [the intermediate function](https://github.com/tlambert03/pymmcore-plus/pull/111/files#diff-dfc1820f65568699d9e948aff4a5dea71ab530d23305903b03dee87db58e88ecR242-R244) that only calls `some_callback` if `propertyChanged` is emitted with the exact device label and property label.

tlambert03:
that said... `property.changed.connect(some_callback)` is so appealing to me too :joy: lemme see what I can finagle

tlambert03:
nevermind... wasn't so bad afterall.  Now it's `property.valueChanged.connect(some_callback)`.  There is a caveat here, which is that if the DeviceProperty itself get's deleted and garbage collected, then anything connected to `property.valueChange` will no longer be triggered (unlike the previous solution, in which the connection would outlive the DeviceProperty object).  thoughts on the relative merits?

ianhi:
ooooh that lack of persistence is a real a bummer.
So I really like the idea of the user facing callback connection API being the same everywhere, but I also think I would be surprised if that callback gets dropped. Which is leading me towards liking your intermediate function with a dummy object for the syntax the best.

I think that that intermediate function could live on the core (possibly publicly - as it's useful there as well) and then the intermediate object would be something like this:

```python
class valueChanged:
    def __init__(self, device: str, label: str, mmc: CMMCorePlus) -> None:
        self._mmc = mmc
        self._dev = device
        self._label = label
    def connect(self, func):
        self._mmc._connect_specific_property_callback(self._dev, self._label, func)
    def disconnect(self, func):
        ...
```
with `self.valueChanged = valueChanged(self.name, self.label, self.core)` in the DeviceProperty init.



So ranking the options:

1. Intermediate function with object for syntax
2. original implementation here
3. consistent connect api but drops the callback on garbage collection
   - The more I think about it the more I'm worried about the potential of this surprising a user someday.

tlambert03:
The real challenge here is 
1) avoiding holding strong references to the provided callback object (particularly if it's a bound method)
while also...
2) allowing the user to call `disconnect()` with the original callback.

psygnal does quite a bit of work to make sure that no strong references to callback objects are maintained.  The real difficulty here is that intermediate function:

```python
def connect(self, callback):
    def _wrapper(device, label, new_value)
        if device == self.device and label == self.name:
            callback(new_value)  # this closure of callback prevents garbage collection of 
    self._mmc.propertyChanged.connect(_wrapper)
```

problem comes with:

```python
obj = SomeBigCostlyClass()
property.valueChanged.connect(obj.some_method)
del obj  # queue for garbage collection ... but it won't work here
```

psygnal does all that, but I'm still trying to figure out how to offer this signal-forwarding behavior without duplicating much logic

tlambert03:
ok, I moved the logic to core, added a `[dis]connectPropertyChangeCallback` method.  Let me know what you think

tlambert03:
> ooh one final thought - should those new methods live on the `events` object or does it make more sense on core?

open to it... but what would you have the API be?  options could be:
```
core.events.property('device', 'label').connect(
core.propertyEvent('device', 'label').valueChanged.connect(...)
```

etc...  thoughts?

ianhi:
oh boy that's a tough choice. Having it on `events` definitely feels the most consistent from a user perspective.

but I'm torn between

```
core.events.property('device', 'label').connect()
core.events.property('device', 'label').changed.connect()
```

can we overload `core.events.propertyChanged`?

So that all of:
```
core.events.propertyChanged().connect
core.events.propertyChanged("device").connect
core.events.propertyChanged("device", "label").connect
```

are possible? Or have I gotten lost in sauce and lost touch with what's actually reasonable

tlambert03:
> Or have I gotten lost in sauce and lost touch with what's actually reasonable

:joy:

it would be cool I agree. but `core.events.propertyChanged("device", "label")` is already an alias for `core.events.propertyChanged.emit("device", "label")`... so we can't do that

tlambert03:
let's leave out the middle `.changed.` object for now, so something like `core.events.property('device', 'label').connect()`.  but it could also be ``core.events.specificPropertyChanged('device', 'label').connect()`.?

ianhi:
I think i've been thinking about this too much and am losing touch with my inner new user that I try to embody when thinking about these things :(

ianhi:
> core.events.specificPropertyChanged('device', 'label').connect()? (if we want the word "changed" in there somewhere)...


ooooh yeah I think that that's the one. That's the only one so far that feels like it doesn't need a docstring

tlambert03:
> I think i've been thinking about this too much and am losing touch with my inner new user that I try to embody when thinking about these things :(

hah. ok. let's not overthink it.  pick:

core.events.property('device', 'label').connect()
core.events.specifiPropertyChanged('device', 'label').connect()

tlambert03:
specificPropertyChanged it is!

ianhi:
Doesn't have to be now, but we can also consider the sibling signal: `devicePropertyChanged("device")`

tlambert03:
> Doesn't have to be now, but we can also consider the sibling signal: devicePropertyChanged("device")

sure, but what core event would that correlate with?

ianhi:
> sure, but what core event would that correlate with?

I'm not sure I understand? I was imaginging doing a similar filtering of `propertyChanged` and  passing it along if the `device` matches, rather than both the device and label

tlambert03:
Ah, I see: so it would be emitted when any property on a specific device is changed.  I like

tlambert03:
in that case, I think perhaps we shouldn't use `events.specificPropertyChanged()` but rather `events.devicePropertyChanged(device: str, property: Optional[str]=None)`?

ianhi:
> in that case, I think perhaps we shouldn't use events.specificPropertyChanged() but rather events.devicePropertyChanged(device: str, property: Optional[str]=None)?

:+1: all the options for users presented in the most consistent possible manner. love it

tlambert03:
that was exhilarating!

ianhi:
> that was exhilarating!

:tada:  :) I think we ended up at a really good place - very exciting!

ianhi:
amazing!

PR #114, opened 2022-03-10
fix flaky tests

ianhi:
Used to actually use `core.run_mda` but it seems like it impossible to get proper timing to not be inconsistent in CI.

Once this passes I will restart a few times to make sure i actually fixed it.


ianhi:
ok that's 3 consecutive passes. I think all is well now.

tlambert03:
Thanks!

PR #115, opened 2022-03-12
Add a new SequenceAcquisition  event signals

fdrgsp:
This PR adds three new `SequenceAcquisition` event signals:

`startContinuousSequenceAcquisition =  Signal()` # when continuousSequenceAcquisition is started
`startSequenceAcquisition = Signal(str, int, float, bool)`  # when SequenceAcquisition is started
`stopSequenceAcquisition = Signal(str)` # when (Continuous)SequenceAcquisition is stopped

and three methods to the core to implement the above signals:
- `startContinuousSequenceAcquisition()`
- `startSequenceAcquisition()`
- `stopSequenceAcquisition()`

linked to PR https://github.com/tlambert03/napari-micromanager/pull/148


ianhi:
oooh actually what do you think about adding this directly to MMCore? I'lll ask over there as well.

ianhi:
See https://github.com/micro-manager/mmCoreAndDevices/issues/179 - even if they do get added there we will need to make some changes here. 

Having looked at that little more carefully I also think we probably can't get away with just one signal here. I think we need both a started and stopped signal are necessary as we should have the parameters of the startedsequence emitted in the started signal.

We also need to be careful here to maintain the typing overloads on startSequenceAcquistion:

https://github.com/micro-manager/pymmcore/blob/5fb6ad14a57d2b10b57d04bc7706ff56f1162c9c/pymmcore/__init__.pyi#L924-L931

which should also get re-implemented here to as they will emit different signals than the `continuous` variant. Difference are  somewhat described here: https://github.com/micro-manager/mmCoreAndDevices/issues/171#issuecomment-1058312997 also helpful to look directly at the implementations:

[`startSequenceAcquisition`](https://github.com/micro-manager/mmCoreAndDevices/blob/356db0164e4678533e02f4ba00e044047d758c2a/MMCore/MMCore.cpp#L2758-L2830)
[`startContinuousSequenceAcquisition](https://github.com/micro-manager/mmCoreAndDevices/blob/356db0164e4678533e02f4ba00e044047d758c2a/MMCore/MMCore.cpp#L2905-L2935)

tlambert03:
plus... tests :)

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -659,6 +659,62 @@ def getImage(self, *args, fix=True) -> np.ndarray:
...
+    def startSequenceAcquisition(self, *args) -> None:
+        if len(args) == 3:
+            numImages, intervalMs, stopOnOverflow = args
+            super().startSequenceAcquisition(numImages, intervalMs, stopOnOverflow)
+            cameraLabel = super().getCameraDevice()
+        else:
+            cameraLabel, numImages, intervalMs, stopOnOverflow = args
+            super().startSequenceAcquisition(
+                cameraLabel, numImages, intervalMs, stopOnOverflow
+            )
```
ianhi, line 695: I think this simplifies it a bit and allows the first thing that happens to be the C++ to check the arguments so we are garunteed to have either 3 or 4, previously that wasn't garunteed and you end up with an error when unpacking args in the else branch (at least I think...)

```suggestion
        super().startSequenceAcquisition(numImages, intervalMs, stopOnOverflow)
        if len(args) == 3:
            numImages, intervalMs, stopOnOverflow = args
            cameraLabel = super().getCameraDevice()
        else:
            cameraLabel, numImages, intervalMs, stopOnOverflow = args
```

ianhi: woops sorry this sholda been `super().startSequenceAcquisition(*args)`


and same deal with stop acquisition

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -659,6 +659,55 @@ def getImage(self, *args, fix=True) -> np.ndarray:
...
+        self.events.startContinuousSequenceAcquisition.emit()
+
+    @overload
+    def startSequenceAcquisition(
+        self,
+        numImages: int,
+        intervalMs: float,
+        stopOnOverflow: bool,
+    ) -> None:
+        ...
```
tlambert03, line 674:
```suggestion
        ...  # pragma: no cover
```

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -659,6 +659,55 @@ def getImage(self, *args, fix=True) -> np.ndarray:
...
+
+    @overload
+    def startSequenceAcquisition(
+        self,
+        cameraLabel: str,
+        numImages: int,
+        intervalMs: float,
+        stopOnOverflow: bool,
+    ) -> None:
+        ...
```
tlambert03, line 684:
```suggestion
        ... # pragma: no cover
```

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -659,6 +659,55 @@ def getImage(self, *args, fix=True) -> np.ndarray:
...
+    @overload
+    def stopSequenceAcquisition(self, cameraLabel: str) -> None:
+        ...
+
+    def stopSequenceAcquisition(self, *args) -> None:
+        """Stop a SequenceAcquisition."""
+        super().stopSequenceAcquisition(*args)
+        cameraLabel = args[0] if args else super().getCameraDevice()
+        self.events.stopSequenceAcquisition.emit(cameraLabel)
+
```
tlambert03, line 710:
```suggestion
    def stopSequenceAcquisition(self, cameraLabel: Optional[str] = None) -> None:
        """Stop a SequenceAcquisition."""
        if cameraLabel is None:
            super().stopSequenceAcquisition()
        else:
            super().stopSequenceAcquisition(cameraLabel)
        cameraLabel = cameraLabel or super().getCameraDevice()
        self.events.stopSequenceAcquisition.emit(cameraLabel)
```

pymmcore_plus/core/events/_qsignals.py
```diff
@@ -26,6 +26,15 @@ class QCoreSignaler(QObject):
...
     mdaEngineRegistered = Signal(object, object)  # new engine, old engine
+    startContinuousSequenceAcquisition = (
+        Signal()
+    )  # when continuousSequenceAcquisition is started
+    startSequenceAcquisition = Signal(
+        str, int, float, bool
+    )  # when SequenceAcquisition is started
+    stopSequenceAcquisition = Signal(
+        str
+    )  # when (Continuous)SequenceAcquisition is stopped
```
tlambert03, line 37:
```suggestion
    # when continuousSequenceAcquisition is started
    startContinuousSequenceAcquisition = Signal()
    # when SequenceAcquisition is started
    startSequenceAcquisition = Signal(str, int, float, bool)
    # when (Continuous)SequenceAcquisition is stopped
    stopSequenceAcquisition = Signal(str)
```

PR #120, opened 2022-03-29
add `setContext` method

ianhi:
A reasonable extension to this (possibly to be included here) is to also allow setProperty to work. Currently it doesn't as this doesn't allow for multiple arguments.


ianhi:
@tlambert03 if you have a chance could you look at this?

basic usage is like this:

```python
core = CMMCorePlus.instance()
with core.setContext(autoShutter = False):
    assert not core.getAutoShutter()
    # do other stuff
assert core.getAutoShutter()
```

tlambert03 (approved):
LGTM

PR #122, opened 2022-04-15
Ensure import of QtWidgets in `check_qt_app`

ianhi:
fixes: https://github.com/tlambert03/pymmcore-plus/issues/119

also cleaned up some unused things in `_util.py`

### DIFF
--- a/pymmcore_plus/_util.py
+++ b/pymmcore_plus/_util.py
@@ -1,3 +1,4 @@
+import importlib
 import os
 import sys
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


ianhi:
Not really sure how to test this automatically as that would require messing with import order inside pytest. But this no longer fails running:
`python -c  "from napari.layers import Points;from pymmcore_plus import CMMCorePlus;core = CMMCorePlus.instance()"`

tlambert03 (approved):

PR #124, opened 2022-05-01
update setShutterOpen() method

fdrgsp:
minor change in the `setShutterOpen()` method:
 - fix annotations
 - use `propertyChanged()` event  instead of `shutterSet` event. 

This is used in https://github.com/tlambert03/napari-micromanager/pull/161.


ianhi:
@fdrgsp does autoshutter need the same?

Also, if we are no longer firing the `shutterSet` event then it should be removed from the events object as well.

fdrgsp:
> @fdrgsp does autoshutter need the same?

the autoshutter only needs the `autoShutterSet` signal that we have already implemented.

> Also, if we are no longer firing the `shutterSet` event then it should be removed from the events object as well.

True, I just removed it!

tlambert03 (approved):

PR #126, opened 2022-05-11
Fix waiting for mda events

ianhi:
yikes. The indentation got messed up at some point when playing with MDA stuff and resulted in us never actually waiting until the time of the next mda event, instead just immediately executing it.

New test fails without the fix.

@tlambert03 would you be open to me making a release after this merges? If so how to do, just push a new tag?

### DIFF
--- a/pymmcore_plus/_tests/test_mda.py
+++ b/pymmcore_plus/_tests/test_mda.py
@@ -1,0 +1,22 @@
+import time
+
+from useq import MDASequence
+
+from pymmcore_plus import CMMCorePlus
+
+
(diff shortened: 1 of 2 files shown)


tlambert03 (approved):
Thank you!  will merge and release

tlambert03:
@ianhi, before I release, can you take a look at the increase in intermittent test failures?  I can't tell if it might have been caused by this PR ... but a number of tests are now randomly failing:
https://github.com/tlambert03/pymmcore-plus/runs/6393500156?check_suite_focus=true
https://github.com/tlambert03/pymmcore-plus/runs/6394080322?check_suite_focus=true

PR #131, opened 2022-06-06
fix: Make sure to set xy in mda even if they are 0

ianhi:
previously the check wasn't working correctly if one of x or y was zero, then they would be incorrectly left in the current position.


tlambert03 (approved):
LGTM

PR #133, opened 2022-07-08
fix: make mda engine exit gracefully after an exception

ianhi:
Fixes https://github.com/tlambert03/pymmcore-plus/issues/130

### DIFF
--- a/pymmcore_plus/mda/_engine.py
+++ b/pymmcore_plus/mda/_engine.py
@@ -248,20 +248,31 @@ class MDAEngine(PMDAEngine):
         sequence : MDASequence
             The sequence of events to run.
         """
-        self._prepare_to_run(sequence)
+        try:
+            self._prepare_to_run(sequence)
 
-        for event in sequence:
-            cancelled = self._wait_until_event(event, sequence)
+            for event in sequence:
+                cancelled = self._wait_until_event(event, sequence)
 
-            # If cancelled break out of the loop
-            if cancelled:
-                break
+                # If cancelled break out of the loop
+                if cancelled:
+                    break
 
-            logger.info(event)
-            self._prep_hardware(event)
+                logger.info(event)
+                self._prep_hardware(event)
 
-            self._mmc.snapImage()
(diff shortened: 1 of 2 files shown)


pymmcore_plus/mda/_engine.py
```diff
@@ -248,20 +248,31 @@ def run(self, sequence: MDASequence) -> None:
...
-        self._finish_run(sequence)
+                self._events.frameReady.emit(img, event)
+            self._finish_run(sequence)
+        except Exception as e:  # noqa E722
+            # clean up so future MDAs can be run
+            self._canceled = False
+            self._running = False
+            try:
+                self._finish_run(sequence)
+            except:  # noqa E722
```
tlambert03, line 275:
```suggestion
            except Exception:
```

pymmcore_plus/mda/_engine.py
```diff
@@ -248,20 +248,31 @@ def run(self, sequence: MDASequence) -> None:
...
+                # ignore this one and raise the first failure
+                pass
```
tlambert03, line 277:
```suggestion
            with contextlib.suppress(Exception):
                self._finish_run(sequence)
```

ianhi: pretty slick.

pymmcore_plus/mda/_engine.py
```diff
@@ -248,20 +249,28 @@ def run(self, sequence: MDASequence) -> None:
...
-            self._events.frameReady.emit(img, event)
+                self._events.frameReady.emit(img, event)
+        except Exception as e:  # noqa E722
+            # clean up so future MDAs can be run
+            self._canceled = False
+            self._running = False
+            with contextlib.suppress(Exception):
+                self._finish_run(sequence)
+            raise e
         self._finish_run(sequence)
```
ianhi, line 276: moved this to below the `try/except` so that if it throws an error we don't then try to throw an error twice.

tlambert03 (approved):
Lgtm!

PR #139, opened 2022-09-05
feat: add a new signals and methods for groups and presets + test

fdrgsp:
This PR adds the following signals:
 - `configGroupDeleted`
 - `configDeleted`
 - `configDefined`

This PR adds the following methods:
 - `deleteConfig()`-> delete a group-preset configuration (emit `configDeleted `)
 - `deleteConfigGroup()` -> delete a group configuration (emit `configGroupDeleted `)
 - `defineConfig()`-> create a group-preset configuration (emit `configDefined `)

example of use: https://github.com/pymmcore-plus/pymmcore-widgets/pull/48.

### DIFF
--- a/pymmcore_plus/core/_mmcore_plus.py
+++ b/pymmcore_plus/core/_mmcore_plus.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import atexit
 import os
 import re
-import warnings
 import weakref
 from contextlib import contextmanager
 from datetime import datetime
@@ -56,7 +55,7 @@ _CHANNEL_REGEX = re.compile("(chan{1,2}(el)?|filt(er)?)s?", re.IGNORECASE)
 STATE = pymmcore.g_Keyword_State
 LABEL = pymmcore.g_Keyword_Label
 STATE_PROPS = (STATE, LABEL)
-UNNAMED_PRESET = 'NewPreset'
+UNNAMED_PRESET = "NewPreset"
 
 
 @contextmanager
@@ -725,7 +724,7 @@ class CMMCorePlus(pymmcore.CMMCore):
             shutterLabel = super().getShutterDevice()
             state = args
         self.events.propertyChanged.emit(shutterLabel, "State", state)
-    
+
     def deleteConfig(self, group: str, preset: str) -> None:
         super().deleteConfig(group, preset)
         self.events.presetDeleted.emit(group, preset)
@@ -733,13 +732,13 @@ class CMMCorePlus(pymmcore.CMMCore):
     def deleteConfigGroup(self, group: str) -> None:
         super().deleteConfigGroup(group)
         self.events.groupDeleted.emit(group)
-    
+
     def deletePresetDeviceProperties(
         self, group: str, preset: str, device_property_list: List[Tuple[str, str]]
     ) -> None:
         """
         Delete the specified list of (device, property) from the specified preset.
-        
+
         This method will emit the "newGroupPreset" signal with group and preset info.
 
         Parameters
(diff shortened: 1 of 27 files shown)


pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -724,6 +725,115 @@ def setShutterOpen(self, *args):
             state = args
         self.events.propertyChanged.emit(shutterLabel, "State", state)
 
+    def deleteConfig(self, group: str, preset: str) -> None:
+        super().deleteConfig(group, preset)
+        self.events.presetDeleted.emit(group, preset)
```
tlambert03, line 730: even if the wording of micro-manager's API is sometimes confusing, I feel like we should stick strictly to it.

So, if someone is wondering what event might get emitted when the `deleteConfig` method is used, I think their first guess would be `configDeleted`

and `deleteConfigGroup` -> `configGroupDeleted`
etc...

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -724,6 +725,115 @@ def setShutterOpen(self, *args):
...
+        for d, p, v in dev_prop_val:
+            if (d, p) not in device_property_list:
+                _to_emit.append((d, p, v))
+                super().defineConfig(group, preset, d, p, v)
+
+        self.events.newGroupPreset.emit(group, preset, _to_emit)
+
+    def defineConfigGroup(self, group: str) -> None:
+        super().defineConfigGroup(group)
+        self.events.newGroup.emit(group)
```
tlambert03, line 767: `configGroupDefined`

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -724,6 +725,115 @@ def setShutterOpen(self, *args):
...
+                super().defineConfigGroup(
+                    group
+                )  # needed to refresh pymmcore 'ChannelGroup' options
+
+            if not preset:
+                idx = sum(UNNAMED_PRESET in p for p in self.getAvailableConfigs(group))
+                preset = f"{UNNAMED_PRESET}_{idx}" if idx > 0 else UNNAMED_PRESET
+
+            super().defineConfig(group, preset)
+            self.events.newGroupPreset.emit(group, preset, [])
```
tlambert03, line 799: this should be `configDefined` ... and it should be emitted in _both_ cases for consistency.  But you could make the arguments `None` in the case if the single argument overload

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -724,6 +725,115 @@ def setShutterOpen(self, *args):
...
+
+    def defineConfigFromDevicePropertyValueList(
```
tlambert03, line 801: are you making this method to avoid event excess emission?

this feels more like something that an end-user should do (remember, keeping the API surface minimal is desirable)... and they can always use event blockers and/or event paused from psygnal if they want to group together event emission

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -724,6 +725,51 @@ def setShutterOpen(self, *args):
...
+    def defineConfig(self, group: str, preset: str, *args) -> None:
+
+        if not preset:
+            idx = sum(UNNAMED_PRESET in p for p in self.getAvailableConfigs(group))
+            preset = f"{UNNAMED_PRESET}_{idx}" if idx > 0 else UNNAMED_PRESET
+
+        if not self.isGroupDefined(group):
+            super().defineConfigGroup(
+                group
+            )  # needed to refresh pymmcore 'ChannelGroup' options
```
tlambert03, line 760:
```python
            # needed to refresh pymmcore 'ChannelGroup' options
            super().defineConfigGroup(group)
```

tlambert03 (approved):

PR #140, opened 2022-09-18
feat: New engine protocol

tlambert03:
Hey @ianhi, @fdrgsp, curious to get your thoughts on this variant of the MDAEngine.

One issue with the current pattern is that it puts everything on the engine, with basically a single run function. As a result, the engine must implement all the "control" stuff as well, like `is_running/paused`, `cancel/toggle_pause`, and it's up to the engine to do all the event emission (not unlike the device adapter pattern in MM, where the events you actually receive as a listener depend totally on which device you're using).  As a result, if we want to improve stuff about the way the engine works in the run function (for instance, to add image processing hooks, etc...) it means that _every_ implementation out there would have to also update itself... making the "extensibility" not all that useful.

This PR creates a new object called `MDARunner` that "runs" an engine, and exposes a single external api for running and pausing, and handles all the event emission.  Now the MDAEngine is only responsible for `setup/teardown_sequence`, `setup/teardown_event`, and `exec_event`.
(We could optionally also make it that if the engine _has_ a run method it gets used, but that it is discouraged?), and extending the pattern with pre/post event hooks would just mean updating our internal Runner object

This is a breaking change of course, so we should try to see if anyone was actually using the previous pattern with github code search (or if you know off hand?)

### DIFF
--- a/pymmcore_plus/mda/_engine.py
+++ b/pymmcore_plus/mda/_engine.py
@@ -1,12 +1,14 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, NamedTuple
 
 from useq import MDAEvent, MDASequence
 
 from ._protocol import PMDAEngine
 
 if TYPE_CHECKING:
+    import numpy as np
+
     from ..core import CMMCorePlus
 
 
@@ -43,4 +45,8 @@ class MDAEngine(PMDAEngine):
     def exec_event(self, event: MDAEvent) -> Any:
         # TODO: add non-aquisition event-specific logic here later
         self._mmc.snapImage()
-        return self._mmc.getImage()
+        return EventPayload(image=self._mmc.getImage())
+
+
+class EventPayload(NamedTuple):
+    image: np.ndarray
--- a/pymmcore_plus/mda/_engine.py
+++ b/pymmcore_plus/mda/_engine.py
@@ -1,214 +1,32 @@
-import contextlib
-import time
-from abc import abstractmethod
-from typing import TYPE_CHECKING, Protocol, runtime_checkable
+from __future__ import annotations
 
-from loguru import logger
+from typing import TYPE_CHECKING, Any
+
 from useq import MDAEvent, MDASequence
 
-from .events import PMDASignaler, _get_auto_MDA_callback_class
+from ._protocol import PMDAEngine
 
 if TYPE_CHECKING:
     from ..core import CMMCorePlus
 
 
-@runtime_checkable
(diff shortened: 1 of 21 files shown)


ianhi:
Big :+1: on the concept from me, this is clearly more extensible. Haven't had time to look at the details closely yet, but I can try to convert my raman-mda-engine to this in the next few days to get familiar.

> As a result, the engine must implement all the "control" stuff as well, like is_running/paused, cancel/toggle_pause
> it means that every implementation out there would have to also update itself... making the "extensibility" not all that useful.


Huge problems indeed. thanks for tackling eliminating them!


> This is a breaking change of course, so we should try to see if anyone was actually using the previous pattern with github code search (or if you know off hand?)

A quick search seems to imply that I was the only user. But there may be more as implied by https://github.com/pymmcore-plus/napari-micromanager/issues/179 which did not show up in a code search

ianhi:
Indeed this gave some nice simplifications here: https://github.com/ianhi/raman-mda-engine/pull/4

but the TODO comment about return object is definitely important as I wasn't quite sure what I should be doing with the spectra I collect.

pymmcore_plus/mda/_engine.py
```diff
@@ -220,57 +38,9 @@ def _prep_hardware(self, event: MDAEvent, waitForSystem: bool = True):
...
-            self._canceled = False
-            self._running = False
-            with contextlib.suppress(Exception):
-                self._finish_run(sequence)
-            raise e
-        self._finish_run(sequence)
+    def exec_event(self, event: MDAEvent) -> Any:
+        # TODO: add non-aquisition event-specific logic here later
+        self._mmc.snapImage()
+        return self._mmc.getImage()
```
ianhi, line 46: Can we specify a subclassable `MDAEventData` object? Then we could have an `image` or maybe `datatype` attributes. That way we can still support images, but also allow for other data types like my use case of collecting spectral data

tlambert03: I like this idea, but what would you think about making it Protocol based rather than class based?

For instance, our internal implementation would just check for certain attributes on the object
```
image = getattr(obj, 'image', None)
```

Someone could then use a `NamedTuple`, or a `dataclass`  ... or a custom object.

Not 100% sure here on the implementation, but I definitely like the idea of returning a container object rather than a numpy array

tlambert03: I changed this to returning a named tuple, with an attribute of `image`.  That allows the `_runner` logic to check for an `image` attribute (and only emit it if it's there)... and we can expand on that protocol later in the future.

Sound good? 

If so, how would you feel about this PR going in soon?

PR #143, opened 2022-09-20
feat: new signals to handle pixel size configurations

fdrgsp:
This PR adds new signals when a new pixel size configuration is `defined` (`definePixelSizeConfig()`) and `deleted` (`deletePixelSizeConfig()`) and a new signal when a px size is `set` (`setPixelSizeUm()`).

Example of use: https://github.com/pymmcore-plus/pymmcore-widgets/pull/25


tlambert03:
each new signal we add hurts me a little bit :joy:

... so I just want to double check that there's definitely no way to get this info from core (i.e. core doesn't emit any more generic property changed signal that we could use here?)

tlambert03:
also, can this be a single signal that says "something about the pixel sizes changed... go look for yourself", rather than three new signals?

fdrgsp:
@tlambert03 
> core doesn't emit any more generic property changed signal that we could use here

I tried to follow the `propertychanged` signal but it is not emitted when a new px size cfg is defined or deleted (`mmc.definePixelSizeConfig()`, `mmc.deletePixelSizeConfig()`) nor when a px size is assigned to a cfg (`mmc.setPixelSizeUm()`)

> also, can this be a single signal that says "something about the pixel sizes changed... go look for yourself", rather than three new signals?

Since the process to setup a new px size cfg in micromanager requires a 2 step process (first define and then assign a px size - `mmc.definePixelSizeConfig()` + `mmc.setPixelSizeUm()`), I thought we needed to separate the 2 different signals...but maybe we can do something different!

tlambert03:
> > each new signal we add hurts me a little bit 😂
> Why?😂

- a small API surface area is always nicer than a large one
- it's always easier stuff than to remove it

so... each time we add something we're making it: 
a) more complicated
b) harder to just look at "all" the events quickly and make quick sense of it
c) harder to quickly see which event might be the best one for a particular case.

etc...

that said, if something is literally impossible without it, as we've seen before, then it's a necessary evil.  but evil it is, and should be done as a last resort :)

tlambert03:
> Since the process to setup a new px size cfg in micromanager requires a 2 step process (first define and then assign a px size - mmc.definePixelSizeConfig() + mmc.setPixelSizeUm()), I thought we needed to separate the 2 different signals...but maybe we can do something different!

can you show me an example of where you're consuming these signals?  (i.e. in widgets?)

pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -724,6 +724,33 @@ def setShutterOpen(self, *args):
...
+    ) -> None:
+        ...
+
+    def definePixelSizeConfig(self, *args) -> None:
+        if len(args) == 1:
+            resolutionID = args
+            super().definePixelSizeConfig(resolutionID)
+        else:
+            resolutionID, deviceLabel, propName, value = args
+            super().definePixelSizeConfig(resolutionID, deviceLabel, propName, value)
```
tlambert03, line 751:
```suggestion
super().definePixelSizeConfig(*args)
```

PR #147, opened 2022-10-02
fix: fix test error after psygnal update

tlambert03:
small fix for the recent psygnal update

### DIFF
--- a/pymmcore_plus/mda/_engine.py
+++ b/pymmcore_plus/mda/_engine.py
@@ -4,6 +4,7 @@ from abc import abstractmethod
 from typing import TYPE_CHECKING, Protocol, runtime_checkable
 
 from loguru import logger
+from psygnal import EmitLoopError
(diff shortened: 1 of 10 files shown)


pymmcore_plus/mda/_engine.py
```diff
@@ -265,7 +266,9 @@ def run(self, sequence: MDASequence) -> None:
                 self._mmc.snapImage()
                 img = self._mmc.getImage()
 
-                self._events.frameReady.emit(img, event)
+                with contextlib.suppress(EmitLoopError):
```
tlambert03, line 269: @ianhi  ... this was added from https://github.com/tlambert03/psygnal/pull/115 ... is this the behavior you wanted here?

tlambert03:
merging so tests can pass again.  but @ianhi, let me know if you envisioned something different for catching callback errors in run_mda

PR #160, opened 2022-11-19
docs: update documentation (using mkdocs)

tlambert03:
@fdrgsp and I would like to work on expanding the docs, probably unifying docs for all repos in the pymmcore-plus org

I have really enjoyed/preferred working with mkdocs in other projects, so this PR is a work-in-progress to create a mkdocs base for pymmcore-plus docs


tlambert03:
ok @ianhi  and @fdrgsp, this is ready for review.  I think it's looking great!

You can see the build preview here: https://pymmcore-plus--160.org.readthedocs.build/en/160/

tlambert03:
one thing I'll point out, in the [table here](https://pymmcore-plus--160.org.readthedocs.build/en/160/api/cmmcoreplus/#cmmcoreplus-api-summary), i've labeled the methods as inherited, reimplemented, or new to CMMCorePlus.  Id like the docstrings for each of those "reimplemented" methods to have a standard little "reason for reimplementing" blurb

fdrgsp (approved):
@tlambert03 This is awesome! Love it!

I just have one comment, what do you think about mentioning/describing the use `CMMCorePlus.instance()` in the `Overview` section? I know that it might be more relevant if used in combination with the `pymmcore-widgets` but I think it is a great feature of `pymmcore-plus`.

tlambert03:
it's in there... do you mean more explicitly point it out?

fdrgsp:
> it's in there... do you mean more explicitly point it out?

Yes, I was thinking to add a more explicit explanation of the difference between using `CMMCorePlus.instance()` vs just `CMMCorePlus()`. I know that probably the people who will use this package will know what a `singleton` is (as written in the `Basic Usage` example) but maybe it is worth mentioning it.

tlambert03:
added!

tlambert03:
@ianhi, this has obviously turned into a huge PR to review (most of the lines are actually docstring improvements as opposed to infrastructure), but I admit I'm pretty attached to the general structure at this point :joy:

I'd like to merge this sooner than later, but don't want to do this one without your sign-off

After this is in, I'd like create similar docs for useq-schema, napari-micromanager, and widgets (which @fdrgsp already has been working on) and host them all together on github pages to make them all appear like they're subpages on the same site.

Will probably merge in a day or so if I don't hear any major objections from you first :)

PR #167, opened 2022-11-27
build: support python 3.11 

tlambert03:
adds tests for py3.11


tlambert03:
that was easy :)  working well

PR #172, opened 2022-12-04
feat: add mock_sample decorator for ease of testing/demonstration

tlambert03:
This adds a `mock_sample` decorator that wants a generator function that yields numpy arrays.
It returns a context manager that, when entered, patches `core.getImage` to return the next image from the generator instead.

```python
from typing import Iterator, Tuple

import numpy as np
from pymmcore_plus import CMMCorePlus, mock_sample

core = CMMCorePlus()


# decorate a function that yields numpy arrays with @mock_sample
@mock_sample(mmcore=core, loop=True)
def noisy_sample(shape: Tuple[int, int] = (10, 10)) -> Iterator[np.ndarray]:
    yield np.random.random(shape)


# use it as a context manager.
# each time `core.getImage()` is called, a new image is yielded from the generator
with noisy_sample():
    for _ in range(3):
        print(core.snap().shape)
```


ianhi:
I really like having an officially method supported to do this patching. I'd like it even better if we could also figure a way to (probably optionally) allow the image generators to take in information about the stage positions. That would allow things like https://mda-simulator.readthedocs.io/en/latest/ to work properly. Alternatively the imagegenerator could make use of the core object but that feels a bit like leakage.

tlambert03:
Yay! You're alive 😊

Wasn't sure if we'd lost you (though I saw your general GitHub contributions have been slow the last few months). Good to see you! 

Yeah, I like that idea and thought  about it a bit.  We could take advantage of the generator send pattern, but I avoided it here; since technically the decorator is declaring the core it wants to patch, it means the function already has access to core if it wants it, no?

tlambert03:
@ianhi  ... what would you think of an api like:

```python
def snap_image(event: useq.MDAEvent) -> np.ndarray:
    ...
```

really, that just be a custom MDAEngine:

```python
class MySimulator:
    def setup_sequence(self, sequence: MDASequence) -> None:
        ...
        ...
    def exec_event(self, event: MDAEvent) -> object:
        img = _generate_image_for(event)
        return EventPayload(img)

class EventPayload(NamedTuple):
    image: np.ndarray

```


I guess the difference would be that we're not actually querying the real `mmcore.getXYPosition()` or anything... we're just asking the simulator to give us the payload for a given event.

So, we could basically make a thin wrapper on the `PMDAEngine` protocol that makes `setup_sequence` and `setup_event` optional rather than required.  and perhaps adds a context manager?

tlambert03:
>  Does it even need to wrap the Engine protocol though?

nope!  it doesn't "need" to ... I was just observing that we already kinda have a protocol for a thing that executes events and returns an image.  I was thinking that might be a way to implement it under the hood... but technically all we need is any old function that accepts an event and returns an array.  wouldn't even need to be a class method.

I kinda like both APIs honestly.  The generator for something that doesn't care at all about positions or channels, and the function that accepts an event for anyone that wants to get fancy.    `@mock_sample` could certainly detect whether it's dealing with a generator function or a regular callable... but do you think that would be API overload?  (if so, we could have `sample_generator` and `mock_mda_sample` or whatever you like)

ianhi:
> I kinda like both APIs honestly.

Very much agree with this!

> @mock_sample could certainly detect whether it's dealing with a generator function or a regular callable... but do you think that would be API overload?

This detection would be my preference, it's basically two options so not a serious overload. The second option is it's own kind of overload in the sense of proliferation of functions with different names that do effectively the same thing.

Thinking a bit about the signatures, should it be that the image shape gets passed to the generator, or rather that the generator defines the image shape and we modify the core img widht and height? The second has the advantage that it nods towards the fact that in a sense what we are doing is making a fake a camera device.

Have to leave for group coffee time so this is slighlty incomplete thoughts but is the core of what I feel

tlambert03:
> This detection would be my preference, it's basically two options so not a serious overload. The second option is it's own kind of overload in the sense of proliferation of functions with different names that do effectively the same thing.

so, I just sat down to try to implement this... and while it's straightforward enough conceptually, from a `typing.overload` perspective it's very much not simple.  Essentially, I'm not sure it's possible to have the _return_ value of a function be a decorator that returns an overloaded function, and that's what we need here (since `mock_sample` has to return the overloaded decorator).  I don't think I explained that well... but i think we need to have two functions

```python
# when decorating a generator function, the decorator returns a
# callable that accepts the same arguments as the generator function,
# and returns a context manager.
@overload
def mock_sample_generator(
    func: Callable[_P, Iterator[np.ndarray]]
) -> Callable[_P, _MockSampleContextManager]:
    ...

# when decorating a regular function that accepts an MDAEvent,
# the decorator returns a callable that takes NO arguments and
# returns a context manager
@overload
def mock_mda_sample(
    func: Callable[[MDAEvent], np.ndarray]
) -> Callable[[], _MockSampleContextManager]:
    ...
```

you ok with that?  at least as a start?

ianhi:
> when decorating a regular function that accepts an MDAEvent,
> the decorator returns a callable that takes NO arguments and
> returns a context manager

I'm not sure I'm fully understanding this, why does the returned function not take any arguments? I think you probably have this right, but I'm getting a bit lost in how decorators work. 


Also to clarify, the use case could look something like this?

```python
class mySimulation:
     def __init__(self, various_params...):
           # set up simulation

    @mock_mda_sample
    def simulated_camera(self, MDAEvent) -> np.ndarray:
          # draw from my simulation

sim = mySimulator()

with sim.simulated_camera():
     core.snap()
     # live mode etc..
```

tlambert03:
> why does the returned function not take any arguments? 

well there's plenty of ways it could work of course... but I'm imagining that the main api is as you had it:

```python
@mock_mda_sample
def simulated_camera(event: MDAEvent) -> np.ndarray:  ...
``

i.e. the decorator takes a `Callable[[MDAEvent], np.ndarray]`.  (whether that callable is a bound class method or a plain function doesn't particularly matter here).

The return value of `mock_mda_sample` then is a callable that takes no arguments, and returns a context manager that, when entered, will ensure that all mda camera snaps inside the context call the original `simulated_camera` function with the current event.  Again, exactly as you had it:

```python
with simulated_camera():  # no arguments here
    ...
```

in other words 

```python
my_context = simulated_camera()
with my_context:
    ...
```

main point being, once decorated the name `simulated_camera` now points not to the original function that takes an MDAEvent, but to a new function that can be called to get an instance of a context manager.

by contrast, with the image yielding pattern:

```python
@mock_sample
def simulated_sample(*args, **kwargs) -> Iterator[np.ndarray]: ...
```

`mock_sample` takes  a `Callable[P, Iterable[np.ndarray]]`.  (where `P` is a `ParamSpec`... meaning we don't really care _what_ arguments `simulated_sample` takes... it's up to the user to decide what's important.  me using `shape` in my example above was just an example.  it really doesn't matter.  All we care about there is that the generator yields numpy arrays when we ask for it. 

So here, the `mock_sample` decorator returns a new callable that takes the same arguments as the original function, and  returns a context manager:

```python
with simulated_camera(*args, **kwargs):  # initialize the generator
    core.snap()  # get first yield
    core.snap()  # get second yield
```

PR #178, opened 2022-12-06
docs: add favicon

ianhi:
I took the png from the github org, ran it through https://favicon.io/favicon-converter/ and took the `.ico` file resulting in this:
image


tlambert03:
lovely! :)

PR #180, opened 2022-12-16
feat: emit channelGroupChanged signal when calling setChannelGroup

fdrgsp:
`pymmcore` `setChannelGroup` currently does not have/emit any signal when is called. 

This PR adds core `channelGroupChanged.emit()` to the `setChannelGroup` method.

### DIFF
--- a/src/pymmcore_plus/core/_mmcore_plus.py
+++ b/src/pymmcore_plus/core/_mmcore_plus.py
@@ -1559,6 +1559,14 @@ class CMMCorePlus(pymmcore.CMMCore):
             args = (super().getCameraDevice(),) + args
         self.events.roiSet.emit(*args)
 
+    def setChannelGroup(self, channelGroup: str) -> None:
+        """Specifies the group determining the channel selection.
+        
+        ...and send a propertyChanged signal.
+        """
+        super().setChannelGroup(channelGroup)
+        self.events.propertyChanged.emit("Core", "ChannelGroup", channelGroup)
+
     def state(self, exclude: Iterable[str] = ()) -> StateDict:
         """Return `StateDict` with commonly accessed state values.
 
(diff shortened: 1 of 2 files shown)


tlambert03 (commented):
Let's check that the value actually changes and only then emit a signal

fdrgsp:
> Let's check that the value actually changes and only then emit a signal

Something like this?
```
    def setChannelGroup(self, channelGroup: str) -> None:
        """Specifies the group determining the channel selection.
        
        ...and send a propertyChanged signal.
        """
        super().setChannelGroup(channelGroup)
        if self.getChannelGroup() != channelGroup:
            return
        self.events.propertyChanged.emit("Core", "ChannelGroup", channelGroup)
```

Also in `pymmcore` there is already a `channelGroupChanged` signal. Do you think it is better to use this instead of the `propertyChanged` one (`self.events.channelGroupChanged.emit(channelGroup))`?

tlambert03:
> Something like this?

I _think_ that won't work... because by the time you check `self.getChannelGroup() != channelGroup:` it will always be the same (cause you just set it to be the same)... right?  I'm thinking something like:

```python
if self.getChannelGroup() != channelGroup:
    super().setChannelGroup(channelGroup)
    self.events.propertyChanged.emit("Core", "ChannelGroup", channelGroup)
```
(alternatively, you could collect `self.getChannelGroup()` _before_ calling the setter, and then make the comparison afterwards).

> Also in pymmcore there is already a channelGroupChanged signal. Do you think it is better to use this instead of the propertyChanged one (self.events.channelGroupChanged.emit(channelGroup))?

oh definitely!  If there is a `channelGroupChanged` signal, then I would certainly expect a `setChannelGroup` method to emit that signal

tlambert03 (approved):

PR #199, opened 2023-06-30
[WIP] feat: add autofocus logic to the engine

fdrgsp:
This PR adds to the `MDAEngine` the logic to control any Micromanager Autofocus device (e.g. Nikon PFS).

Linked to https://github.com/pymmcore-plus/useq-schema/pull/91.

(test will fail until useq-schema PR is merged)

### DIFF
--- a/src/pymmcore_plus/mda/_engine.py
+++ b/src/pymmcore_plus/mda/_engine.py
@@ -40,28 +40,18 @@ class MDAEngine(PMDAEngine):
         event : MDAEvent
             The event to use for the Hardware config
         """
-        # TODO:
-        # - if len(event.sequence.positions) == 1, sequence.z_plan is relative and autofocus is engaged:
-        #     - calculate z start from current position
-
-        # if only 1 position, and z_plan is relative, and autofocus is engaged,
-        # calculate z from current position
-        if (
-            len(event.sequence.stage_positions) == 1
-            and not event.z_is_autofocus
-            and not isinstance(event.sequence.z_plan, NoZ())
-            and self._mmc.isContinuousFocusLocked()
-        ):
-            z_pos = self._mmc.getPosition(event.z_device or self._mmc.getFocusDevice())
-        else:
-            z_pos = event.z_pos
-
         if event.x_pos is not None or event.y_pos is not None:
             x = event.x_pos if event.x_pos is not None else self._mmc.getXPosition()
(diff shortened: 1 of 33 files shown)


src/pymmcore_plus/mda/_engine.py
```diff
@@ -32,6 +58,10 @@ def setup_sequence(self, sequence: MDASequence) -> None:
 
         self._mmc = self._mmc or CMMCorePlus.instance()
 
+        # switch off autofocus device if it is on
+        with contextlib.suppress(RuntimeError):
+            self._mmc.setProperty(self._mmc.getAutoFocusDevice(), "State", "Off")
```
ianhi, line 63: why not:

`self._mmc.setContinuousFocus(False)`

src/pymmcore_plus/mda/_engine.py
```diff
@@ -59,6 +115,18 @@ def exec_event(self, event: MDAEvent) -> Any:
...
+        """Perform the autofocus.
+
+        Returns the correction to be applied to the focus motor position.
+        """
+        self._mmc.setPosition(
+            autofocus_event.autofocus_z_device_name,
+            cast(float, autofocus_event.af_motor_offset),
+        )
+        self._mmc.fullFocus()
+        return self._mmc.getZPosition() - cast(float, autofocus_event.z_stage_position)
```
ianhi, line 128: I've found that there absolutely need to be some waits in here. either `waitForSystem` or `waitForDevices`.

On some of my older systems i've also found it be significantly more robust to catch an error in autofocus and retry once or twice. Anecdotally on the olympus system I use 1 retry works most of the time, and 2 retries has yet to fail even over a 16 hour timelapse.

Here's the equivalent section of my engine:

https://github.com/ianhi/raman-mda-engine/blob/f7238bc286aa9642331eb5c6215c458e31327115/raman_mda_engine/_engine.py#L273-L284

ianhi: > I've found that there absolutely need to be some waits in here.

without the waits on some systems the `getZPosition` may not report on the correctly focused stage position.

src/pymmcore_plus/mda/_engine.py
fdrgsp: @ianhi you meant `self._mmc.enableContinuousFocus(False)`? Sure, it should work as well!

tlambert03:
closing because I think #204 is preferred now?

PR #202, opened 2023-07-08
feat: add build-dev command to build dev dependencies on apple silicon

tlambert03:
with this PR, devs with apple silicon should be able to run

```
mmcore build-dev
```

to get working demo camera and utilties


tlambert03:
cc @ashesh-0 and @fdrgsp  ... this should be an easier way to get native demo drivers for testing without rosetta

PR #204, opened 2023-07-15
feat: update engine with autofocus using user Action

### LINKED ISSUE #51: Autofocus in mda
Cross linking with https://github.com/pymmcore-plus/useq-schema/issues/4

Ideally this would support Hardware autofocus. But in addition we can build in an opportunity for software autofocus, e.g. having a signal emitted in MDA after the hardware is set up. Something similar to the pycromanager `post_hardware_hook` (https://pycro-manager.readthedocs.io/en/latest/apis.html?highlight=post_hardware#pycromanager.Acquisition)

fdrgsp:
This pr is a followup of #199 but uses the new `MDAEvent` `Action` logic for hardware autofocus: https://github.com/pymmcore-plus/useq-schema/pull/94.

closes #199
closes #51

Linked to https://github.com/pymmcore-plus/useq-schema/pull/96.


src/pymmcore_plus/mda/_engine.py
```diff
@@ -44,21 +51,65 @@ def setup_event(self, event: MDAEvent) -> None:
...
         self._mmc.waitForSystem()
 
-    def exec_event(self, event: MDAEvent) -> Any:
-        """Execute an individual event and return the image data."""
-        # TODO: add non-aquisition event-specific logic here later
-        self._mmc.snapImage()
-        return EventPayload(image=self._mmc.getImage())
+    def exec_event(self, event: MDAEvent) -> tuple[ndarray, MDAEvent] | None:
+        """Execute an individual event."""
+        action = event.action if hasattr(event, "action") else AcquireImage()
```
tlambert03, line 70: change this to `action = getattr(event, 'action', None)` ... then don't create an empty `AcquireImage()` thing... just keep the same bit at the end that will hit of none of the other action instance checks hit 
```py
self._mmc.snapImage()
return EventPayload(image=self._mmc.getImage())
```

src/pymmcore_plus/mda/_engine.py
```diff
@@ -44,21 +51,65 @@ def setup_event(self, event: MDAEvent) -> None:
...
             self._mmc.setExposure(event.exposure)
 
         self._mmc.waitForSystem()
 
-    def exec_event(self, event: MDAEvent) -> Any:
-        """Execute an individual event and return the image data."""
-        # TODO: add non-aquisition event-specific logic here later
-        self._mmc.snapImage()
-        return EventPayload(image=self._mmc.getImage())
+    def exec_event(self, event: MDAEvent) -> tuple[ndarray, MDAEvent] | None:
```
tlambert03, line 68: why are you changing the signature of the return here?

src/pymmcore_plus/mda/_protocol.py
```diff
@@ -35,7 +35,7 @@ def setup_event(self, event: MDAEvent) -> None:
         """
 
     @abstractmethod
-    def exec_event(self, event: MDAEvent) -> object:
+    def exec_event(self, event: MDAEvent) -> Any:
```
tlambert03, line 38: same here... why change?

src/pymmcore_plus/mda/_runner.py
```diff
@@ -144,17 +146,21 @@ def run(self, sequence: MDASequence) -> None:
...
 
-                if (img := getattr(output, "image", None)) is not None:
-                    with contextlib.suppress(EmitLoopError):
-                        self._events.frameReady.emit(img, event)
+                # emit frameReady signal
+                if output is not None:
+                    img, ev = output
+                    if isinstance(img, ndarray) and isinstance(ev, MDAEvent):
+                        with contextlib.suppress(EmitLoopError):
+                            self.events.frameReady.emit(img, ev)
```
tlambert03, line 163: why change?

src/pymmcore_plus/mda/_engine.py
```diff
@@ -54,11 +67,51 @@ def setup_event(self, event: MDAEvent) -> None:
...
+
+        Returns the new z focus position.
+        """
+        self._mmc.setPosition(
+            action.autofocus_device_name,
+            action.autofocus_motor_offset,
+        )
+        self._mmc.waitForSystem()
+
+        # perform fullFocus 3 times in case of failure
```
tlambert03, line 98: I think this should be an option on the action actually:
`max_retries: int = 3` ...

fdrgsp: You mean an option of the `useq.HardwareAutofocus` `Action` like this?

```py
class HardwareAutofocus(Action):
    """
    ...
    max_retries : int
        The number of retries if autofocus fails.
    """

    type: Literal["hardware_autofocus"] = "hardware_autofocus"
    autofocus_device_name: str
    autofocus_motor_offset: float
    max_retries: int
```

and then modify this as:
```py
 # perform fullFocus 3 times in case of failure
 for i in range(action.max_retries):
     try:
         self._mmc.fullFocus()
         self._mmc.waitForSystem()
     except RuntimeError:
         warnings.warn(f"Hardware autofocus failed {i} times.", stacklevel=2)

 return self._mmc.getZPosition()
```

tlambert03: yep, exactly, but max_retries can take a default of 3, and you'll need to fix the try loop, cause this one will *always* perform `fullFocus` 3 times, regardless of success or failure, and the the previous one only retried in the case of failure.  It's a fun little challenge: i.e. how can you code a nested try catch with a variable number of retries? :)

fdrgsp: That's true. This should work:
```py
# perform fullFocus 'action.max_retries' times in case of failure
        for i in range(action.max_retries):
            try:
                self._mmc.fullFocus()
                self._mmc.waitForSystem()
                return self._mmc.getZPosition()

            except RuntimeError:
                warnings.warn(
                    f"Hardware autofocus failed {i + 1} time(s), "
                    f"({action.max_retries - (i + 1)} retries remaining).",
                    stacklevel=2,
                )
        warnings.warn(
            "Hardware autofocus failed. Maximum retries exceeded.", stacklevel=2
        )
        return self._mmc.getZPosition()
```

tlambert03: yep!  minor suggestions:
I don't think it needs a warning on each try... so you can just use:
```python
with suppress(RuntimeError):
    ...
```

and then if it fails, rather than `warnings.warn`. it should come from the logger

fdrgsp: should be ok now!

tlambert03: nice!

tests/test_mda.py
```diff
@@ -98,6 +98,116 @@ def cb(img, event):
...
+    )
+    with qtbot.waitSignals(
+        [
+            core.mda.events.frameReady,
+            core.mda.events.sequenceFinished,
+        ]
+    ):
+        core.run_mda(mda)
+
+    assert core.getPosition() == 100
```
tlambert03, line 119: can you add a comment here to explain what this is testing?  My only question upon looking at it is whether this pretty much just ends up testing that mock_fullfocus ran?  or whether it's testing something deeper about the autofocus process?

src/pymmcore_plus/mda/_engine.py
```diff
@@ -54,11 +69,46 @@ def setup_event(self, event: MDAEvent) -> None:
...
+        action = getattr(event, "action", None)
+
+        # execute hardware autofocus
+        if isinstance(action, HardwareAutofocus) and event.z_pos is not None:
+            # get position index
+            p_idx = event.index.get("p", None)
+            # run autofocus and get the new z position
+            new_z = self._execute_autofocus(action)
+            # get the correction to apply to each z position
+            self._z_correction[p_idx] = new_z - event.z_pos
```
tlambert03, line 82:
```suggestion
            # update the correction for this position
            self._z_correction[p_idx] = self._execute_autofocus(action)
```
i think it would read _slightly_ clearer if `self._execute_autofocus()` returned a delta, ie. it should return the `new_z - event.z_pos` rather than just the `new_z`.  And to do so, it should use `self._mmc.getPosition()` inside of the body of `_execute_autofocus` rather than using `event.z_pos` here (which is correct of course, but only insofar as `setup_event()` *elsewhere* actually set the position to event.z_pos.  So this is just a bit more explicit

fdrgsp: I used the `event.z_pos` mainly because I thought that in case of a relative z-stack, we are more precise if we use the event info instead of the stage position since we want the slice interval to be precisely what we set in the `MDASequence`. But maybe it really doesn't matter and is the same. Do you mean something like this:

```py
def exec_event(self, event: MDAEvent) -> Any:
        """Execute an individual event."""
        action = getattr(event, "action", None)

        # execute hardware autofocus
        if isinstance(action, HardwareAutofocus) and event.z_pos is not None:
            # get position index
            p_idx = event.index.get("p", None)
            # run autofocus and get the correction to apply to each z position
            self._z_correction[p_idx] = self._execute_autofocus(action)

            return None

        # acquire an image and emit the image data
        self._mmc.snapImage()
        return EventPayload(image=self._mmc.getImage())

    def _execute_autofocus(self, action: HardwareAutofocus) -> float:
        """Perform the hardware autofocus.

        Returns the z correction to apply to each z position.
        """
        self._mmc.setPosition(
            action.autofocus_device_name,
            action.autofocus_motor_offset,
        )
        self._mmc.waitForSystem()
        
        current_z = self._mmc.getZPosition()

        @retry(exceptions=RuntimeError, tries=action.max_retries, logger=logger.warning)
        def _full_focus() -> float:
            self._mmc.fullFocus()
            self._mmc.waitForSystem()
            return self._mmc.getZPosition() - current_z

        try:
            return _full_focus()
        except RuntimeError:
            logger.warning("Warning: Hardware autofocus failed.")
            return self._mmc.getZPosition()
```

src/pymmcore_plus/mda/_engine.py
```diff
@@ -32,6 +39,9 @@ def setup_sequence(self, sequence: MDASequence) -> None:
 
         self._mmc = self._mmc or CMMCorePlus.instance()
 
+        # switch off autofocus device if it is on
+        self._mmc.enableContinuousFocus(False)
```
tlambert03, line 43: is there ever a case where we wouldn't want to turn this off?  Like, what about a multi-channel timelapse at a single XYZ position.  If the user had it on to begin with, shouldn't we just leave it on?

src/pymmcore_plus/mda/_engine.py
```diff
@@ -2,8 +2,11 @@
 
 from typing import TYPE_CHECKING, Any, NamedTuple
 
-from useq import MDAEvent, MDASequence
+from useq import HardwareAutofocus, MDAEvent, MDASequence  # type: ignore
```
tlambert03, line 5: why the type ignore?

src/pymmcore_plus/mda/_engine.py
```diff
@@ -54,11 +69,46 @@ def setup_event(self, event: MDAEvent) -> None:
...
+            return None
+
+        # acquire an image and emit the image data
         self._mmc.snapImage()
         return EventPayload(image=self._mmc.getImage())
 
+    def _execute_autofocus(self, action: HardwareAutofocus) -> float:
+        """Perform the hardware autofocus.
+
+        Returns the z correction to apply to each z position.
```
tlambert03, line 90:
```suggestion
        Returns the change in ZPosition that occurred during the autofocus event. 
```

src/pymmcore_plus/mda/_engine.py
```diff
@@ -54,11 +69,46 @@ def setup_event(self, event: MDAEvent) -> None:
...
+        def _full_focus() -> float:
+            self._mmc.fullFocus()
+            self._mmc.waitForSystem()
+            return self._mmc.getZPosition() - current_z
+
+        try:
+            return _full_focus()
+        except RuntimeError:
+            logger.warning("Warning: Hardware autofocus failed.")
+            return 0.0
```
tlambert03, line 110: you could also put this try/except above in `exec_event`

```python
try:
    self._z_correction[p_idx] = self._execute_autofocus(action)
except RuntimeWarning:
    logger.warning("Warning: Hardware autofocus failed.")
```

which would prevent you from overriding a previous correction (that might indeed be closer to the truth) with the `0.0` value

src/pymmcore_plus/mda/_engine.py
```diff
@@ -44,8 +54,13 @@ def setup_event(self, event: MDAEvent) -> None:
             x = event.x_pos if event.x_pos is not None else self._mmc.getXPosition()
             y = event.y_pos if event.y_pos is not None else self._mmc.getYPosition()
             self._mmc.setXYPosition(x, y)
+
         if event.z_pos is not None:
-            self._mmc.setZPosition(event.z_pos)
+            p_idx = event.index.get("p", None)
+            if p_idx not in self._z_correction:
+                self._z_correction[p_idx] = 0.0
+            self._mmc.setZPosition(event.z_pos + self._z_correction[p_idx])
```
tlambert03, line 62:
```suggestion
            correction = self._z_correction.setdefault(p_idx, 0.0)
            self._mmc.setZPosition(event.z_pos + correction)
```

src/pymmcore_plus/mda/_engine.py
fdrgsp: I did some tests with the Nikon PFS and if the PFS is on before running     `self._mmc.setPosition(action.autofocus_device_name, action.autofocus_motor_offset)`, it does not precisely go to the correct position. This is why I added this line.

tlambert03: right, so, what I'm saying is that it clearly does make sense in cases where you _know_ you are going to use autofocus at some point.  But here you've disabled autofocus *all* the time, regardless of anything.  So, if a user had manually turned on autofocus, and hadn't asked for any autofocus in their MDASequence, wasn't moving positions or anything, this would still turn it off on them.  That doesn't seem right?

fdrgsp: oh that's true! Ok so we can move this inside the `exec_event` if the event is an autofocus event I guess!

tlambert03: it's certainly fine to have it here too, but you need to guard with some kind of conditional at least

fdrgsp: I want to do few test on the microscope...

PR #205, opened 2023-07-16
refactor: relax signature of `MDARunner.run` to `Iterable[MDAEvent]`

tlambert03:
This changes the signature of MDARunner.run from 

```py
    def run(self, sequence: MDASequence) -> None:
```

to 

```py
    def run(self, events: Iterable[MDAEvent]) -> None:
```

as discussed elsewhere, this is the more powerful approach, and doesn't tie the engine to sequences of known length/shape, etc...  So, it could accept _any_ custom sequence/generator/iterable of `useq.MDAEvent` (which includes instances of `MDASequence`.)

The only tricky bit were the `setup_sequence` callback and `sequenceStarted` signal.  To avoid breaking uses like @ianhi 's [engine](https://github.com/ianhi/raman-mda-engine/blob/f7238bc286aa9642331eb5c6215c458e31327115/raman_mda_engine/_engine.py#L214) ... When a non `MDASequence` object is passed to `run()`, a mock `GeneratorMDASequence` is created and passed to those signals. It will warn and do nothing if someone tries to iterate over it or check its length/shape, etc...  That object is also exported at the top level so one can check `isinstance(obj, GeneratorMDASequence)`  if they'd like

@ianhi how does this look to you?

### DIFF
--- a/src/pymmcore_plus/__init__.py
+++ b/src/pymmcore_plus/__init__.py
@@ -22,11 +22,13 @@ from .core import (
     PropertyType,
 )
 from .core.events import CMMCoreSignaler, PCoreSignaler
+from .mda._runner import GeneratorMDASequence
 
 if TYPE_CHECKING:
     from .remote import RemoteMMCore, server
 
 __all__ = [
+    "__version__",
     "ActionType",
     "CMMCorePlus",
     "CMMCoreSignaler",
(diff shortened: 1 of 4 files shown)


tlambert03:
I'm pretty sure that this is the direction I want to go in here, so I'm going to merge this.  @ianhi if you have thoughts later and need/want changes, feel free to open an issue.  I  don't think this should be breaking anything for you, just make it easier to express more complicated sequences without using MDASequence

PR #206, opened 2023-07-16
feat: Add support for hardware sequencing

tlambert03:
This adds support for hardware sequencing/merging of events:

1. The `PMDAEngine` Protocol gains a new optional method: `event_iterator`, which can be used to wrap the `for event in self._engine. event_iterator(events):` in the `MDARunner.run()` method.  (if an engine doesn't provide it, it simply uses `for event in iter(events)`
2. our default `MDAEngine` uses this method to combine events, using the new  `CMMCorePlus.canSequenceEvents` method (this is a long method that checks whether two events are sequence able).  If it finds events that can be merged, it merges them into a new `SequencedEvent` type, otherwise it yields the regular event
3. both `setup_event` and `exec_event` event on the Engine now need to be able to handle the Sequenced event.  I have a basic `setup_event` implementation, but haven't added exec support yet.

@fdrgsp ... would be great to have you also work on this PR and try to test it on s15


tlambert03:
note, tests are currently broken because we assert that certain events are emitted, which is no longer the case if events are being merged

tlambert03:
ok, this is all set to go, and I'll probably merge it later today unless @fdrgsp has any reservations

tlambert03:
(we still need to determine what the protocol will be for signaling that a sequence has occurred... for now, it just does the laborious thing of collecting and emmiting all the images)

PR #207, opened 2023-07-17
feat: add general retry utility

tlambert03:
Since it's a common thing to retry stuff a couple times in MMCore, this adds a utility to help with it:

```python
from pymmcore_plus._util import retry

@retry(exceptions=RuntimeError, delay=0.5, logger=print)
def autofocus(core) -> float:
    core.fullFocus()
    core.waitForSystem()
    return core.getZPosition()
```


tlambert03:
cc @fdrgsp for #51

ianhi:
very cool!

PR #208, opened 2023-07-18
refactor: Remove remote feature

tlambert03:
getting rid of remote... i'm sure it doesn't work anymore


tlambert03:
so long remote,  we barely knew ye

PR #215, opened 2023-07-24
Add set_mda to use custom MDARunner

wl-stepp:
Hey everyone,

I would like to use pymmcore-plus with a slightly modified runner that allows for addition of acquisition events. For this I would have to register a different runner in the CMMCorePlus instance. Do you think it makes sense to add this functionality?
@tlambert03 

Thanks, Willi

### DIFF
--- a/src/pymmcore_plus/core/_mmcore_plus.py
+++ b/src/pymmcore_plus/core/_mmcore_plus.py
@@ -1343,6 +1343,20 @@ class CMMCorePlus(pymmcore.CMMCore):
         """
         return self._mda_runner
 
+    def set_mda(self, mda) -> None:
+        """Set the `MDARunner` for this `CMMCorePlus` instance.
+
+        :sparkles: *This method is new in `CMMCorePlus`.*
+
+        Parameters
+        ----------
+        mda : MDARunner
+            Any object conforming to the `MDARunner` protocol.
+        """
+        self._mda_runner = mda()
+        self._mda_runner.set_engine(MDAEngine(self))
+
+
     def run_mda(self, events: Iterable[MDAEvent], block: bool = False) -> Thread:
         """Run a sequence of [useq.MDAEvent][] on a new thread.
 
--- a/src/pymmcore_plus/core/_mmcore_plus.py
+++ b/src/pymmcore_plus/core/_mmcore_plus.py
@@ -1356,7 +1356,6 @@ class CMMCorePlus(pymmcore.CMMCore):
         self._mda_runner = mda()
         self._mda_runner.set_engine(MDAEngine(self))
 
-
     def run_mda(self, events: Iterable[MDAEvent], block: bool = False) -> Thread:
         """Run a sequence of [useq.MDAEvent][] on a new thread.
 
--- a/tests/test_mda.py
+++ b/tests/test_mda.py
@@ -6,8 +6,8 @@ from unittest.mock import Mock, patch
 
 import pytest
 from pymmcore_plus import CMMCorePlus
-from pymmcore_plus.mda.events import MDASignaler
 from pymmcore_plus.mda import MDARunner
+from pymmcore_plus.mda.events import MDASignaler
 from useq import MDAEvent, MDASequence
 
 if TYPE_CHECKING:
@@ -130,8 +130,11 @@ def test_mda_iterable_of_events(
     assert start_mock.call_count == 1
     assert frame_mock.call_count == 2
 
+
 @pytest.mark.parametrize("seq", SEQS)
-def test_mda_replace_runner(core: CMMCorePlus, seq: Iterable[MDAEvent], qtbot: QtBot) -> None:
+def test_mda_replace_runner(
+    core: CMMCorePlus, seq: Iterable[MDAEvent], qtbot: QtBot
+) -> None:
     core.set_mda(MDARunner)
     test_mda_iterable_of_events(core, seq, qtbot)
     test_mda_failures(core, qtbot)
--- a/tests/test_mda.py
+++ b/tests/test_mda.py
@@ -7,6 +7,7 @@ from unittest.mock import Mock, patch
 import pytest
 from pymmcore_plus import CMMCorePlus
 from pymmcore_plus.mda.events import MDASignaler
+from pymmcore_plus.mda import MDARunner
 from useq import MDAEvent, MDASequence
 
 if TYPE_CHECKING:
@@ -128,3 +129,11 @@ def test_mda_iterable_of_events(
 
     assert start_mock.call_count == 1
     assert frame_mock.call_count == 2
+
+@pytest.mark.parametrize("seq", SEQS)
+def test_mda_replace_runner(core: CMMCorePlus, seq: Iterable[MDAEvent], qtbot: QtBot) -> None:
+    core.set_mda(MDARunner)
+    test_mda_iterable_of_events(core, seq, qtbot)
+    test_mda_failures(core, qtbot)
+    test_mda_waiting(core)
+    test_setting_position(core)


tlambert03:
Hey @wl-stepp, @kmdouglass  :wave:  thanks for the PR. 

Thrilled to hear you're looking into using pymmcore-plus.
Let's definitely figure out how we can support your use case!

The goal was to make the engine the user-facing protocol, and not the runner.  If we make the runner mutable, then that creates the possibility that different runners might fail to emit the same events.  While that's certainly not the end of the world (for example, if you're the only one ever using your runner, it doesn't matter if you do or don't emit events that you don't care about), it would be nice if we could keep that level of abstraction and fix your use case another way.

Can you elaborate a bit on your use case?  I feel like two alternate possibilities would be preferable:

1) you could use the `event_iterator` wrapper that I'm adding in https://github.com/pymmcore-plus/pymmcore-plus/pull/206 (it's there to allow for hardware sequencing, but is really much more powerful).  I can explain further, but would like 
2) If what you need is the ability to dynamically insert events using hooks or something, that's always been a planned feature, and if you're at the place where the lack of hooks is limiting you, I'd be happy to talk about getting them in. 
3) as a final fallback possibility, I think rather than a `set_mda` method, I would prefer making the `MDARunner` class type  a class attribute on the `CMMCorePlus` class, so you could overwrite it by subclassing CMMCorePlus and simply changing to your own runner.  Something like:

```py
class CMMCorePlus:
    runner_class = MDARunner

    def __init__(self):
         ...
         self._mda_runner = self. runner_class()
```

Since your lab has as much experience thinking about this as anyone, id certainly be happy to jump on a zoom as well and hear more about your use case, and then put in whatever work we need to support it.

wl-stepp:
Hey @tlambert03,

thanks for the quick response. Yes, I was thinking of testing the implementation of one of our event-driven acquisition workflows in pymmcore-plus. The toggle_pause() function works Ok, but I think it would be much more flexible to just insert acquisition events whenever we want a higher frame rate, or even additional channels etc. I have done something similar using the Micro-Magellan/Java AcquisitionEngine here: [https://github.com/LEB-EPFL/eda_plugin/blob/master/eda_plugin/actuators/pycromanager.py](https://github.com/LEB-EPFL/eda_plugin/blob/master/eda_plugin/actuators/pycromanager.py).

My naive approach would have been to replace the events with a queue and the for loop with a while loop that checks for remaining events.

```python
self.acq_events = deque(events)
while len(self.acq_events) > 0:
    event = self.acq_events.popleft()
```

And then append/insert events in an event-triggered function:

```python
@mmc.mda.events.frameReady.connect
def add_frame(img: np.ndarray, event: MDAEvent):
    if event.global_index > 5:
        return
    new_event = MDAEvent(index={'t': mmc.mda.acq_events[-1].index["t"] + 1, 'c': 0},
                     channel=mmc.mda.acq_events[-1].channel,
                     exposure=50.0,
                     min_start_time= mmc.mda.acq_events[-1].min_start_time + 3.0,
                     global_index=mmc.mda.acq_events[-1].global_index + 1)
    mmc.mda.acq_events.append(new_event)
```

Happy to discuss further what the best way would be to implement something like that.

tlambert03:
> I think it would be much more flexible to just insert acquisition events whenever we want a higher frame rate, or even additional channels etc.

yep, of course.  That's absolutely a top priority, and mostly have been waiting for time and/or someone like you to come along and want the feature.  

> I have done something similar using the Micro-Magellan/Java AcquisitionEngine here

This is very helpful.  It definitely looks like we have a ton of overlapping functionality (e.g. your event_bus object is similar to one of the main motivations for this library, which was to create a qt-like api for connecting callbacks to signals from the core).  And the main difference, if you haven't already figured it out, is that we're trying to keep java and zmq out of the loop entirely here when possible.

> My naive approach would have been to replace the events with a queue and the for loop with a while loop that checks for remaining events.

Yep that would definitely work fine, and I do think that that `event_iterator` method on the engine that I was discussing above could accomplish that for you.  but of course there are many places/levels at which we could attack this

Your code is very helpful, let me look through it today and tomorrow and I'll get back to you with some thoughts and perhaps a zoom meeting?

wl-stepp:
>  And the main difference, if you haven't already figured it out, is that we're trying to keep java and zmq out of the loop entirely here when possible.
Same motivation here.

Sure, happy to zoom about this. Let me know when you had time to have a look and we can fix a meeting.

tlambert03:
summarizing our discussion.  here is a general pattern for subclass the MDAEngine that could be used to inject functionality and control flow at various points:

```py
from queue import Queue
from typing import Any, Iterable, Iterator

from pymmcore_plus.mda import MDAEngine
from useq import MDAEvent, MDASequence


class MyEngine(MDAEngine):
    # you have access to the core here at self.mmcore

    # called once at the beginning of a sequence
    def setup_sequence(self, sequence: MDASequence) -> None:
        return super().setup_sequence(sequence)

    # called before each event in the sequence
        # do pre-setup
        super().setup_event(event)
        # do post-setup

    # called after setup_event to execute
    def exec_event(self, event: MDAEvent) -> Any:
        # do pre-setup
        if "my_key" in event.metadata:
            return self._do_a_thing(event)
        else:
            return super().exec_event(event)
        # do post-exec
        # Note that the runner itself will emit image payloads.
        # but you can also call self.mmcore.getImage() use the return value directly

    # wrapper that controls yielding of events
    def event_iterator(self, events: Iterable[MDAEvent]) -> Iterator[MDAEvent]:
        # you don't have to implement this, but the simplest implementation would simply be:
        yield from events
        return

        # alternatively
        # if you wanted to have some internal _queue in your engine that controlled
        # the yielding of events, this would be a good place for it.
        self._queue = Queue(events)

        # and then somehow determine when to yield events
        while not self._queue.empty():  # or whatever
            yield self._queue.get()

        return  # stop the experiment whenever by returning


core = CMMCorePlus()
# override the default engine with your subclass.
core.mda.set_engine(MyEngine(core))

# build your experiment.
# you can use useq.MDASequence to do this, or just build a list of MDAEvents manually
# or even just start with an empty list and let your event_iterator make them
# all dynamically.
seq: Iterable[MDAEvent] = [...]
core.mda.run(seq)
```

After you spend some time with it, let me know if you'd still like to fully customize the runner (as in this PR), or if you need a bit more control over pausing for `min_start_time` in the `MDARunner._wait_for_event`  method.

tlambert03:
very helpful, thank you!  Yeah that certainly works fine, but I think your call to `iter(self._queue.get, STOP_EVENT)` is the key here.   As long as the *only* thing you're using the `ReaqtEngine` and `mmc.mda.set_engine` for is to implement the `event_iterator` method, you can avoid subclassing MDAEngine and calling `set_engine` altogether.

The key bit is the fact that `CMMCorePlus.run_mda()` expects nothing more than an iterable of `MDAEvents` and ` iter(q.get, STOP_EVENT)` *is* an iterable of `MDAEvents`.

So, distilled down to the key bits, we have:
```python
from queue import Queue

from useq import MDAEvent

# setup core as usual
core = CMMCorePlus.instance()
# this line is necessary for now, but we can add a field to MDAEvent to avoid this.
# will explain in a moment
core.mda.engine.use_hardware_sequencing = False

# now create a queue and a stop sentinel (it doesn't need to be an MDAEvent)
q: Queue[MDAEvent] = Queue()
STOP = object()

# this IS an Iterable[MDAEvent] and can be passed directly to run_mda
event_iterable = iter(q.get, STOP)

# starts a thread and watches the queue for events
core.run_mda(event_iterable)
```

now you can just drop MDAEvents into the queue

```python
some_event = MDAEvent(x_pos=100.0, y_pos=100.0, exposure=10)
q.put(some_event)
```

in action:

https://github.com/pymmcore-plus/pymmcore-plus/assets/1609449/2f623a5f-591a-4bf9-a3e7-5704668d8782

----

**a note about hardware sequencing**

you'll note the line  `core.mda.engine.use_hardware_sequencing = False` in there.  That's necessary for this dummy example right now because, by default, the engine will attempt to *combine* MDAEvents if possible into a single hardware-triggered acquisition.  But in order to do that, it grabs the _next_ event from the iterator.  if we just keep putting simple dummy `MDAEvent()`  objects on the queue, it keeps thinking "yep, I can put *all* of these into one big sequence" and it keeps waiting to see an event that it _can't_ sequence (which would be one at a different position, or a different exposure, etc...).  For your case, you'll probably always just want to set `core.mda.engine.use_hardware_sequencing = False` (and honestly, this example makes me think we should probably set that to False by default, and require opting in to hardware sequencing).  

Alternatively, we could have a flag on the `MDAEvent` object itself that says "execute me immediately, and don't wait to see what else is coming"

kmdouglass:
Thanks @tlambert03 for the feedback! I agree this is very nice and simple way to drive the acquisition via a queue.

I'll speak in person to @wl-stepp later this week, but I'd say this is the beat solution.

tlambert03:
great!  if it's ok with you, I'm going to close this PR, since I think it's safe to say that overriding the runner itself is not immediately necessary

wl-stepp:
Awesome, thank you! Will have a look soon and try to implement this into a 'real' EDA loop with some analysis attached.

PR #219, opened 2023-07-27
feat: add logging to file and cli command

### LINKED ISSUE #108: use `setPrimaryLogFile`
Core has the functions: `[s,g]etPrimaryLogFile` https://micro-manager.org/apidoc/MMCore/latest/class_c_m_m_core.html#a8d929d26376ddc5d57706730c3560d35

it looks as though they default to being empty (though possibly could be set in a config file? I know you can for channel group at least.) in which case logging is disabled.

We should set this to default to somewhere reasonable if it's empty - and then possibly also provide a function for reading from the log.

### LINKED ISSUE #217: TODO: document logging
include logging to file, passing core logs through, log level, etc...

tlambert03:
closes #108
closes #217

### DIFF
--- a/docs/_hooks.py
+++ b/docs/_hooks.py
@@ -18,6 +18,7 @@ HAS_RUN = False  # sentinel to prevent loop on mkdocs serve
 CORE_API_TABLE = "{{ CMMCorePlus_API_Table }}"
 PLUS_MEMBERS = "{{ CMMCorePlus_Members }}"
 CORE_MEMBERS = "{{ CMMCore_Members }}"
+CLI_LOGS = "{{ CLI_Logs }}"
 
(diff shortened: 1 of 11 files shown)


ianhi:
looking back at https://github.com/pymmcore-plus/pymmcore-plus/issues/108 does it also make sense to support/override the mmcore function `setPrimaryLogFile`?

tlambert03:
Will look into it

tlambert03:
ok, I added a call to `setPrimaryLogFile` in the init method now, so all logs go to the log file.  There's a new cli helper `mmcore logs` to view, stream, or reveal, or clear the logs.  and a `configure_logging` method to customize it.

Docs added as well.

PR #233, opened 2023-07-31
bug: add warnings if devices are not found

### LINKED ISSUE #226: Task: Catch case of missing autofocus device
@fdrgsp, with the autofocus stuff in #204 we introduced a stronger possibility that an MDASequence might be run on a system that cannot support it (since it explicitly lists the device name).  Since not every system will have an autofocus device, we should make sure to fail gracefully in that case (perhaps with a warning that AF was requested by cannot be supported).  Can you open a PR to make sure that we don't crash if AF is requested on a scope that can't support it?  You could actually do that more generally for all the things (e.g. if there is no `getXYStageDevice`, etc...)

fdrgsp:
This PR closes #226 and handles warnings in case some of the micromanager devices are not loaded/found.


fdrgsp:
@tlambert03 at the moment I've added warnings on stages and autofocus, do you think we should add warnings for other devices like if channel or cameras are not found?

tlambert03 (approved):
lovely!  thanks!

added a simple logging test to the end of yours

PR #246, opened 2023-08-15
feat: add metadata to frameready signal, add getTaggedImage

### LINKED ISSUE #220: Task: emit metadata along with images during mda
In some way shape or form, we need to be emitting paired metadata along with images during mda.

could be an ome-types object, or a more minimal/faster object that could be used to construct OME metadata later

tlambert03:
This emits a (currently-unstructured) metadata dict as the third value of frameReady.  This would be an important component for a proper writer spec (#29) 

It also adds `getTaggedImage` to mimic the Java API.
i should note that I don't really like the `getTaggedImage` API (the "tags" themselves are constructed in a way that makes them hard to disambiguate).  But for now, this uses it in case it helps with cross-compatibility

side-note: because both psygnal and qt let you connect a slot that accepts less values than the signal emits, we can always add positional arguments to signals without breaking any APIs.

closes #220

### DIFF
--- a/src/pymmcore_plus/core/_mmcore_plus.py
+++ b/src/pymmcore_plus/core/_mmcore_plus.py
@@ -17,6 +17,7 @@ from typing import (
     Callable,
     Iterable,
     Iterator,
+    NamedTuple,
     Pattern,
     Sequence,
     TypeVar,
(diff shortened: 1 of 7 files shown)


tlambert03:
well shoot... looks like this does actually break napari-micromanager after all... need to look into it

PR #248, opened 2023-08-16
fix: fix a bug in saveSystemConfiguration method

fdrgsp:
in #214 we  updated the original `pymmcore` `saveSystemConfiguration` method so that it also saves any pixel size configuration if are set. 

However it is not done properly, it assumes that the objective device e is called "Objective" and it does not take into account any other possible device-property that might be added to the pixel size configuration.

This PR fixes this issue and saves any `device-property-value` that are added to the pixel size `configuration.

---
For example, if we have a px config that is called `cfg1` and we call:
```py
getPixelSizeConfigData(cfg1).dict()
```
we might get something like this: 
``` py
{'Core': {'XYStage': 'XY'}, 'Objective': {'Label': 'Nikon 10X S Fluor'}}
```
Since in the px configuration we have two devices (`Core` and `Objective`), we need to make sure that both will be added to the saved configuration like below:
```
PixelSize,Res10x,Core,XYStage,XY
ConfigPixelSize,Res10x,Objective,Label,Nikon 10X S Fluor
PixelSize_um,Res10x,1.0
PixelSizeAffine,Res10x,1.0,0.0,0.0,0.0,1.0,0.0
```

### DIFF
--- a/src/pymmcore_plus/core/_mmcore_plus.py
+++ b/src/pymmcore_plus/core/_mmcore_plus.py
@@ -1728,28 +1728,29 @@ class CMMCorePlus(pymmcore.CMMCore):
             super().setFocusDevice(focusLabel)
             self.events.propertyChanged.emit("Core", "Focus", focusLabel)
 
-    def saveSystemConfiguration(
-        self, filename: str, objective_device: str | None = None
-    ) -> None:
+    def saveSystemConfiguration(self, filename: str) -> None:
         """Saves the current system configuration to a text file.
 
         **Why Override?** To also save pixel size configurations.
         """
         super().saveSystemConfiguration(filename)
         px_configs = self.getAvailablePixelSizeConfigs()
-        if not px_configs or objective_device is None:
+        if not px_configs:
             return
         # saveSystemConfiguration does not save the pixel size config so here
         # we add to the saved file also any pixel size config.
         with open(filename, "a") as f:
             f.write("# PixelSize settings")
             for px_config in px_configs:
-                data = self.getPixelSizeConfigData(px_config)
-                obj = data.dict()[objective_device]["Label"]
+                data = self.getPixelSizeConfigData(px_config).dict()
+                obj_device = next(iter(data.keys()))
+                obj_label = data[obj_device].get("Label")
+                if obj_label is None:
+                    continue
                 px_size = self.getPixelSizeUmByID(px_config)
                 px_affine = self.getPixelSizeAffineByID(px_config)
                 cfg = (
-                    f"\nConfigPixelSize,{px_config},Objective,Label,{obj}\n"
+                    f"\nConfigPixelSize,{px_config},{obj_device},Label,{obj_label}\n"
                     f"PixelSize_um,{px_config},{px_size}\n"
                     f"PixelSizeAffine,{px_config},{','.join(map(str, px_affine))}"
                 )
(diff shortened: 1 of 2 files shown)


src/pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -1734,24 +1736,32 @@ def saveSystemConfiguration(self, filename: str) -> None:
...
-            for px_config in px_configs:
-                data = self.getPixelSizeConfigData(px_config)
-                obj = data.dict()["Objective"]["Label"]
-                px_size = self.getPixelSizeUmByID(px_config)
-                px_affine = self.getPixelSizeAffineByID(px_config)
-                cfg = (
-                    f"\nConfigPixelSize,{px_config},Objective,Label,{obj}\n"
-                    f"PixelSize_um,{px_config},{px_size}\n"
-                    f"PixelSizeAffine,{px_config},{','.join(map(str, px_affine))}"
+            self._save_pixel_configurations(f)
```
tlambert03, line 1742: this will open and close the file unnecessarily if self.getAvailablePixelSizeConfigs() doesn't return anything.  Also, though it's a private method, technically `TextIOBase` doesn't "demand" that the file have been opened in "append" mode, so the decoupling there leads to the (admittedly tiny) possibility that `_save_pixel_configuration` would overwrite the full context of the file (and not just append).  So, change `_save_pixel_configuration` to take a filename instead of an open handle, and move the open() call inside of that method

tlambert03:
> However it is not done properly, it assumes that the objective device e is called "Objective" and it does not take into account any other possible device-property that might be added to the pixel size configuration.

it seems you came across this as a bug, right?  Can you add a test that would simulate the problem that would fail without your fixes here?

src/pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -1734,24 +1736,32 @@ def saveSystemConfiguration(self, filename: str) -> None:
...
+        px_configs = self.getAvailablePixelSizeConfigs()
+        if not px_configs:
+            return
+        cfg = ["# PixelSize settings"]
+        for px_config in px_configs:
+            data = self.getPixelSizeConfigData(px_config).dict()
+            for device, prop in data.items():
+                cfg.extend(
+                    f"ConfigPixelSize,{px_config},{device},{k},{v}"
+                    for k, v in prop.items()
```
tlambert03, line 1754:
```suggestion
            cfg.extend(
                f"ConfigPixelSize,{px_config},{device},{prop},{val}"
                for device, prop, val in self.getPixelSizeConfigData(px_config)
```

the config object is already an iterable that gives you everything you need.  no need to cast to a dict and then iterate over `items()`

fdrgsp:
> Can you add a test that would simulate the problem that would fail without your fixes here?

I've updated the test so that it will fail without the modification of this PR.

src/pymmcore_plus/core/_mmcore_plus.py
```diff
@@ -1734,24 +1734,30 @@ def saveSystemConfiguration(self, filename: str) -> None:
...
         super().saveSystemConfiguration(filename)
-        px_configs = self.getAvailablePixelSizeConfigs()
-        if not px_configs:
-            return
         # saveSystemConfiguration does not save the pixel size config so here
         # we add to the saved file also any pixel size config.
+        self._save_pixel_configurations(filename)
+
+    def _save_pixel_configurations(self, filename: str) -> None:
         with open(filename, "a") as f:
```
tlambert03, line 1742: :joy: you moved it here, but _still_ opened the file without checking if there are any configs that need to be saved! :)

move the open to where it's needed: just immediately before you write to the file.  line 1760

tlambert03: think of it like opening the refrigerator: you don't open the fridge, then go make coffee, then come back to the fridge to pull out the eggs :)

fdrgsp: yup! make sense :)

tlambert03 (approved):
thanks!

PR #255, opened 2023-08-29
feat:  microscope model, for building configs and in-memory representations of core

tlambert03:
alterative to, and closes #252 

similar to MicroscopeModel in MMStudio.

The general goal here is to have a model  (a set of dataclasses) that contain all of the data in a core instance, while being independent.  There are `create_from...`, `update_from...` and `apply_to...` methods that let you construct models from a current core state, or apply models to a specific core instance.  But, they need not be intrinsically linked to a specific core instance (unlike the other Device/Property objects, which may eventually be replaced by these).  The immediate application for this model is for a config wizard: https://github.com/pymmcore-plus/pymmcore-widgets/pull/183


tlambert03:
going to merge this.  may need a little more iteration, but it's working pretty well both in tests and when i try it in the config wizard.  want to be able to use it in pymmcore-widgets tests as well

PR #262, opened 2023-09-21
feat: add mda thread relay

### LINKED ISSUE #189: Should `MDARunner` manage an `io` thread for usage by anything writing to disk?
I'm thinking about a more general solution to https://github.com/pymmcore-plus/napari-micromanager/pull/258

## Problem

While it's an ok solution for napari-micromanager to spawn a thread in order to update it's zarr, it's not ideal that any MDAWriter would either have to do it's work on the acquisition thread, or to go through the hassle of making it's own io thread. So as it stands there are two options:

1. Writers make their own threads
2. Writers slow down acquisitions

## Solution?

I think that this is a similar issue to what we ran into with `MDAEngine`'s having to manage timing of the next event. We can both save a lot of downstream complexity and encourage better performance on the acquisition thread by having the `MDARunner` (or a similar object) manage an single `io` thread. Downstream consumers of the results of the MDA (i.e. displays and writers) could register their `write_to_disk` functions (hopefully all GIL releasing) to be called from that thread.

```python

class MyWriter:
    def __init__(self, core, ....):
           # connect sequence started/finished signals
    def onSequenceStarted():
       ....
    def onFrame(self, idx: tuple[int,...], event, img):
         # tifffile.imsave(...)
         # zarr.something
         # other

writer = MyWriter(core)

core.mda.io.register_function(writer.onFrame)
```

and the `MDARunner` could place the event output in a cross thread queue here:
https://github.com/pymmcore-plus/pymmcore-plus/blob/5fb239ded2ae7a455dcc9e5afd6fb0ecfcb7db3e/src/pymmcore_plus/mda/_runner.py#L153

and then in the `io` thread draw from queue and do:

```python
event, img = queue.get()
idx = event_to_idx(event)

for f in self._disk_write_fxns:
    f(idx, event, img)
```

tlambert03:
This implements a general pattern similar to what we're doing in napari-micromanager to pass data to handlers (such as writers) asynchronously as it comes off the MDA.

- a new `MDARelayThread` object has a deque
- it collects signals as they come off the `core.mda.events` (such as a `frameReady`) signal
- it runs in another thread and *synchronously* calls anything connected to it
- before the context manager exits, it (by default) makes sure the dequeue is 

This means that anyone can implement a performant data handler such as what @ianhi  and @fdrgsp  did in https://github.com/pymmcore-plus/napari-micromanager/pull/272 without having to worry about implementing the threading.  And multiple handlers can share the same handler thread.  (note: this assumes that the callback doesn't need to update the MDA itself, which is the case for writers, etc...)

so it's:

```mermaid
graph LR;
    core.mda.events -->|frameReady| ThreadRelay;
    ThreadRelay --> Handler1;
    ThreadRelay --> Handler2;
    ThreadRelay --> Handler3;
```

```python
import time

import useq
from pymmcore_plus.mda import mda_listeners_connected

core = CMMCorePlus.instance()


class SlowHandler:
    def frameReady(self, ary, event):
        time.sleep(0.12)
        print("slowly processing frame:", event.index)


seq = useq.MDASequence(time_plan=useq.TIntervalLoops(loops=4, interval=0.1))
with mda_listeners_connected(SlowHandler()):
    core.mda.run(seq)
```

outputs:
```
2023-09-21 09:11:40,484 - pymmcore-plus - INFO - (_runner.py:216) MDA Started: Multi-Dimensional Acquisition with: nt=4
2023-09-21 09:11:40,485 - pymmcore-plus - INFO - (_runner.py:184) index=mappingproxy({'t': 0}) min_start_time=0.0
2023-09-21 09:11:40,586 - pymmcore-plus - INFO - (_runner.py:184) index=mappingproxy({'t': 1}) min_start_time=0.1
slowly processing frame: {'t': 0}
2023-09-21 09:11:40,685 - pymmcore-plus - INFO - (_runner.py:184) index=mappingproxy({'t': 2}) min_start_time=0.2
slowly processing frame: {'t': 1}
2023-09-21 09:11:40,789 - pymmcore-plus - INFO - (_runner.py:184) index=mappingproxy({'t': 3}) min_start_time=0.3
2023-09-21 09:11:40,801 - pymmcore-plus - INFO - (_runner.py:308) MDA Finished: Multi-Dimensional Acquisition with: nt=4
slowly processing frame: {'t': 2}
slowly processing frame: {'t': 3}
```

(note that the slow handler didn't hold up the MDA...).

closes #189 

cc @ianhi 

### Summary by CodeRabbit

- New Feature: Added an optional parameter `qt_connection_type` to the `listeners_connected` function in `src/pymmcore_plus/_util.py`. This allows users to specify the Qt connection type when connecting signals, enhancing flexibility.
- New Feature: Introduced a new module `mda_listeners_connected` in `src/pymmcore_plus/mda/_thread_relay.py`. This provides a context manager for connecting MDA events to listeners and supports both synchronous and asynchronous execution of callbacks, improving performance and usability.
- Test: Added a new test function `test_mda_listeners_connected` in `tests/test_thread_relay.py` to ensure the correct behavior of the new context manager.

### DIFF
--- a/src/pymmcore_plus/_util.py
+++ b/src/pymmcore_plus/_util.py
@@ -334,9 +334,9 @@ def listeners_connected(
     listeners : Any
         Object(s) that has methods matching the name of signals on `emitter`.
     name_map : dict[str, str] | None
-        Optionally map signal names on `emitter` to signal names on `listener`.  This
-        can be used to connect signals with different names. By default, the signal
-        names must match exactly.
+        Optionally map signal names on `emitter` to different method names on
+        `listener`.  This can be used to connect callbacks with different names. By
+        default, callbacks names must match the signal names exactly.
 
     Examples
     --------
--- a/src/pymmcore_plus/_util.py
+++ b/src/pymmcore_plus/_util.py
@@ -15,6 +15,8 @@ import appdirs
 if TYPE_CHECKING:
     from typing import Any, Callable, Iterator, Literal, TypeVar
 
+    QtConnectionType = Literal["AutoConnection", "DirectConnection", "QueuedConnection"]
+
     from typing_extensions import ParamSpec, TypeGuard
 
     from .core.events._protocol import PSignalInstance
@@ -315,7 +317,10 @@ def _sorted_rows(data: dict, sort: str | None) -> list[tuple]:
 
 @contextmanager
 def listeners_connected(
-    emitter: Any, *listeners: Any, name_map: dict[str, str] | None = None
+    emitter: Any,
+    *listeners: Any,
+    name_map: dict[str, str] | None = None,
+    qt_connection_type: QtConnectionType | None = None,
 ) -> Iterator[None]:
     """Context manager for listening to signals.
 
@@ -337,6 +342,12 @@ def listeners_connected(
         Optionally map signal names on `emitter` to different method names on
         `listener`.  This can be used to connect callbacks with different names. By
         default, callbacks names must match the signal names exactly.
+    qt_connection_type: str | None
+        ADVANCED: Optionally specify the Qt connection type to use when connecting
+        signals, in the case where `emitter` is a Qt object.  This is useful for
+        connecting to Qt signals in a thread-safe way. Must be one of
+        `"AutoConnection"`, `"DirectConnection"`, `"QueuedConnection"`.
+        If `None` (the default), `Qt.ConnectionType.AutoConnection` will be used.
 
     Examples
     --------
@@ -380,8 +391,13 @@ def listeners_connected(
         for attr_name in common_names:
             if _is_signal_instance(signal := getattr(emitter, attr_name)):
                 slot_name = name_map.get(attr_name, attr_name)
+                args: tuple[Any, ...] = ()
                 if callable(slot := getattr(listener, slot_name)):
-                    tokens[attr_name].add(signal.connect(slot))
+                    if qt_connection_type and "Qt" in type(signal).__module__:
+                        from qtpy.QtCore import Qt
+
+                        args = (getattr(Qt.ConnectionType, qt_connection_type),)
+                    tokens[attr_name].add(signal.connect(slot, *args))
 
     try:
         yield
--- a/src/pymmcore_plus/_util.py
+++ b/src/pymmcore_plus/_util.py
@@ -391,13 +391,15 @@ def listeners_connected(
         for attr_name in common_names:
             if _is_signal_instance(signal := getattr(emitter, attr_name)):
                 slot_name = name_map.get(attr_name, attr_name)
-                args: tuple[Any, ...] = ()
                 if callable(slot := getattr(listener, slot_name)):
                     if qt_connection_type and "Qt" in type(signal).__module__:
                         from qtpy.QtCore import Qt
 
-                        args = (getattr(Qt.ConnectionType, qt_connection_type),)
-                    tokens[attr_name].add(signal.connect(slot, *args))
+                        ctype = getattr(Qt.ConnectionType, qt_connection_type)
+                        token = signal.connect(slot, ctype)  # type: ignore
+                        tokens[attr_name].add(token)
+                    else:
+                        tokens[attr_name].add(signal.connect(slot))
 
     try:
         yield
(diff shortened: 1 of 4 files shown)


ianhi:
Looks very cool. I actually have a use case that I've been working towards so this is great timing. I'll give this a shot wiht developing that today + tomorrow (post numerous lab meetings ;()

ianhi:
@tlambert03 I see you're still making some changes, is this ready for trying out using, or should I wait?

ianhi:
Just reading over the code here it looks a bit as though the only way to access this is by entering a context manager. Is that correct? What about the use case where you aren't directly initiating the mda? For example if I set up a listener from a jupyter notebook, then run some other code, then start an MDA from within napari-micromanager?

tlambert03:
correct, this is a low-level tool at the moment.

> For example if I set up a listener from a jupyter notebook, then run some other code, then start an MDA from within napari-micromanager?

we can certainly add convenience APIs on top of this in the future.  such as a `core.mda.add_handler` method or something like that.   Then `Runner.run()` would enter/exit the context for the user

ianhi:
> we can certainly add convenience APIs on top of this in the future. such as a core.mda.add_handler method or something like that. Then Runner.run() would enter/exit the context for the user

Maybe I'm missing something but context manager feels like the convenience layer, and the ability to add from anywhere the fundamental. A simpler example than my use might be something like: napari-micro uses this threading to handle it's image display, and separately a user creates a writer. They both need to have access to this, but can't easily both be included in the context manager call

tlambert03:
> context manager feels like the convenience layer

the context manager is there to make sure that nothing stays connected too long.  if someone _didn't_ use a context manager, then these connections would have the ability to persist dangerously past their usefulness.  Perhaps give me the API you'd _like_ to see, and we can discuss how to get it there.  (you don't need to use the context manager if you don't want... but it still makes sure to cleanup things that we definitely want to cleanup)

tlambert03:
> They both need to have access to this, but can't easily both be included in the context manager call

The user doesn't need to use the context:

```python
class Runner:
    def add_handler(self, handler):
        self._handlers.append(handler)
    
    def run():
        with mda_listeners_connect(self._handlers, self):
            ...

class DisplayHandler:
    ...

class WriterHandler:
    ...

core = CMMCorePlus.instance()
core.mda.add_handler(DisplayHandler())
core.mda.add_handler(WriterHandler())
core.mda.run()
```

ianhi:
here your `___Handler` objects are user created and know how to respond to `frameReady`?

tlambert03:
correct.  as shown in the original post, and following from #247, any methods on a handler that match names on the emitter (in this case `core.mda.events`) will be connected on entry and disconnected on exit.  (That's be default, by can be modified using `name_map`: https://github.com/pymmcore-plus/pymmcore-plus/pull/260)

ianhi:
>  the context manager is there to make sure that nothing stays connected too long. if someone didn't use a context manager, then these connections would have the ability to persist dangerously past their usefulness.

In the abstract that makes sense to me, but curious what you view as a connection that could be dangerous. Maybe there's something to be said for adding connections with different expiration times?  Such as ("indefinite", "after next mda", "after 3 events", etc...)



> rom https://github.com/pymmcore-plus/pymmcore-plus/pull/247, any methods on a handler that match names on the emitter

ahh I see, very nice

tlambert03:
> what you view as a connection that could be dangerous

one of the primary bugs we always ran into in napari were connections that outlived the receiver.  So, if the callback is dead for whatever reason, and the emitter emits to it, you'll get an error.   So, it's just "good hygiene" to be mindful of the full lifecycle of connections between signals and slots. 

> Maybe there's something to be said for adding connections with different expiration times? Such as ("indefinite", "after next mda", "after 3 events", etc...)

That could perhaps be added to something like the `add_handler` method i played with in https://github.com/pymmcore-plus/pymmcore-plus/pull/262#issuecomment-1730122424

and could be implemented as a `pop` from `self._handlers` after some predicate is met.

tlambert03:
hey @ianhi, I'll likely merge this PR today.  Absolutely happy (and want) to keep discussing how best to expose this general functionality at a higher level.  I think the underlying logic is sound, and we can continue to decide how best to use it and what layers of convenience to add on top of it

ianhi (approved):
@tlambert03 looks good to me!

PR #265, opened 2023-09-22
feat: OMETiff writer

tlambert03:
also includes #262 

### Summary by CodeRabbit

- New Feature: Added support for multi-dimensional acquisition (MDA) with the `pymmcore_plus` library and saving the acquired data as OME-TIFF files.
- New Feature: Introduced `OMETiffWriter` class for handling the writing of OME-TIFF files, including initialization, sequence start handling, and frame writing.
- New Feature: Added a context manager `mda_listeners_connected` for connecting MDA events to listeners, supporting both synchronous and asynchronous execution of callbacks.
- Test: Added tests for the new `mda_listeners_connected` context manager and `OMETiffWriter` functionality.
- Refactor: Updated `listeners_connected` function in `_util.py` to include an optional parameter `qt_connection_type` for specifying Qt connection type.


src/pymmcore_plus/mda/handlers/_ome_tiff_writer.py
```diff
@@ -0,0 +1,107 @@
...
+            metadata["PhysicalSizeX"] = pix
+            metadata["PhysicalSizeY"] = pix
+            metadata["PhysicalSizeXUnit"] = "µm"
+            metadata["PhysicalSizeYUnit"] = "µm"
+
+        # TODO:
+        # there's a lot we could still capture, but it comes off the microscope
+        # over the course of the acquisition (such as stage positions, exposure times)
+        # ... one option is to accumulate these things and then use `tifffile.comment`
+        # to update the total metadata in sequenceFinished
```
wl-stepp, line 102: Yeah, this would also include real frame times for example, right? For the adaptive acquisitions these are pretty important and depending on the hardware, the TimeIncrement set is not always a very good measure for the real arrival time of frames. Is there a good value for this in the metadata received in frameReady?

tlambert03: Yep exactly. Actual time would be present in the `meta` dict (third argument of frameReady) and would go into the Plane.delta_t field in the ome model.

Side note, see also the merged image writer in #267.  I imagine that one also being useful for you, since it would better handle possible ragged arrays that don't nicely fit into a clean nD model

tlambert03:
some decisions must be made here regarding file and metadata layout, and many of those have been discussed and made multiple times before in other related repos.  here's a helpful comment from @ctrueden at https://lists.openmicroscopy.org.uk/pipermail/ome-devel/2016-April/003619.html


-------------

**TL;DR**: Embed the OME-XML into one or more OME-TIFF headers. Do not rely
solely on a companion file.


For what it's worth, here is how we do multi-file multi-page OME-TIFF
format with WiscScan:

- We write all focal planes for a given timepoint+position+channel to a
single OME-TIFF file.
- Each timepoint, position and channel gets its own OME-TIFF file.

E.g.:
HCx19_Ant_LP3_A1_C<c>_TP<t>_SP<sp>_FW<fw>.ome.tiff

Where:
- '<c>' is the 0-indexed channel number, i.e. the OME C dimension
- '<t>' is the 0-indexed timepoint, i.e. the OME T dimension
- '<sp>' is the 0-indexed stage position, with each constituting a separate
OME Image
- '<fw>' is the 0-indexed filter wheel filter number, when imaging with
multiple filters, with each constituting a separate OME Image

So you end up with a collection of OME-TIFF files described by an OME-XML
block which may span multiple <Image> elements—one per <sp>+<fw>
combination. This scheme is extensible to any number of extra 'dimensions'
although of course the OME-XML won't retain any structural details of how
these various Image elements are related.

As for storing actual OME-XML string:

- We store the full thing into the first (0, 0, ..., 0) TIFF file
ImageDescription (as per the spec [1])

- We store the full thing again into every Nth TIFF file, where N is
configurable but typically 10

- In the other OME-TIFFs, we write an OME-XML stub using the <BinaryOnly>
element—e.g.:

<?xml version="1.0" encoding="UTF-8" standalone="no"?><OME xmlns="
http://www.openmicroscopy.org/Schemas/OME/2012-06" xmlns:xsi="
http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="
http://www.openmicroscopy.org/Schemas/OME/2012-06
https://www.openmicroscopy.org/Schemas/OME/2012-06/ome.xsd"><BinaryOnly
MetadataFile="HCx33_Post_Long_20xObj_x1pt5_int10_p220_g282_xystep383pt4_zstep27_C0_TP0_SP0_FW0.ome.tiff"
UUID="urn:uuid:a14af27e-c6f0-48b4-bb14-e0d4f98925f0"/></OME>

For us, this is a nice compromise between redundancy (in case one or more
TIFF files are lost) and efficiency (since the OME-XML can become several
megabytes or more in size).

We also write out a copy of the OME-XML metadata to a companion file, but
this is for the user's convenience only—it is not used or referenced from
the actual OME-TIFF files. When OME-TIFF was first invented, it was deemed
an important quality that the metadata _not_ be stored only in a separate
companion file, because users would too frequently lose or delete said
metadata file, leaving them in a bad place.

To actually do the work:

- WiscScan uses the Bio-Formats C++ bindings to build up the OME-XML.

- The OME-XML is kept in the Bio-Formats memory structure as it is being
built up.

- It is written out to the temporary companion file as the run proceeds, in
case of a crash.

- WiscScan writes the TIFF binary data using libtiff, leaving a placeholder
stub in ImageDescription for later overwriting with the OME-XML.

- When the OME-XML is finalized (i.e. when the run is complete, and all
metadata is known and populated into the data structure), then the OME-XML
is appended to the relevant TIFF file(s) efficiently, using the Bio-Formats
"overwrite comment" approach [2].

src/pymmcore_plus/mda/handlers/_ome_tiff_writer.py
```diff
@@ -0,0 +1,107 @@
...
+        imwrite(self._filename, shape=shape, dtype=dtype, metadata=metadata)
+
+        # memory map numpy array to data in OME-TIFF file
+        self._mmap = memmap(self._filename)
```
wl-stepp, line 106: I'm playing around with this and ran into a small problem. If I have an acquisition with one axes with length 1, the memmap call swallows that axis.
So this MDASequence for example:
``` python
2023-11-10 10:19:42,768 - pymmcore-plus - INFO - (_runner.py:311) MDA Finished: metadata={'pymmcore_widgets': {'version': '0.4.2.dev58+g3f1e2ea.d20231109'}}
axis_order=('t', 'c') 
channels=(Channel(config='488', group='Channel', exposure=100.0, do_stack=True, z_offset=0.0, acquire_every=1, camera=None),)
time_plan=TIntervalLoops(prioritize_duration=False, interval=datetime.timedelta(microseconds=200000), loops=20)
```
frameReady receives the index for channel however, leading for too many indices for the mmep. I have to do:
``` python
        self._mmap = memmap(self._filename)
        self._mmap = cast("np.memmap", self._mmap)
        self._mmap = self._mmap.reshape(shape)
```

tlambert03: thanks @wl-stepp, this comment was very helpful.  using `mmap.shape = shape` immediately after instantiating the memmap appears to work to reinforce singleton dimensions with the tifffile-created memmap.  Fixed and test added.

tlambert03:
the OME-XML still isn't being written, but since this is the only writer that will write a single 5D tiff (which will be interpreted correctly by bioformats in imageJ), i think it's worth getting in for playing with, and we can flesh out the OMEXML later...

PR #270, opened 2023-09-26
feat: autofind autofocus offset devices

tlambert03:
In micro-manager, there is some variability in the way that autofocus devices are implemented.  Some have a separate offset device, while others can directly set the offset of an associated device.  As a result, calling `setAutoFocusOffset`, may or may not do anything depending on the current autofocus device, and the way to set the offset of these devices is to call `setPosition` on a corresponding Stage Device (see https://github.com/micro-manager/micro-manager/issues/858).

For many devices adapters (such as the NikonTI), it is possible to know *a priori* what the corresponding offset device name would be.  This PR adds a map of known device lookups to offset device names, and sets the corresponding position accordingly.  The goal is to make `setAutoFocusOffset` be *the* way to update the autofocus device offset (rather than sometimes having to call `setPosition`.  Currently, it only adds NikonTE2000 and NikonTI ... but there are a number of commented out offset devices that may or may not be useful to uncomment eventually.

It's a little magical, but it's also unlikely to affect anything but a couple devices like Nikon at the moment.

@marktsuchida curious if you think this is a terrible idea :)

### Summary by CodeRabbit

- New Feature: Introduced `setAutoFocusOffset` method in the core class to apply an offset to the one-shot focusing device, enhancing flexibility in configuring autofocus devices.
- Refactor: Updated `_execute_autofocus` method in the MDA engine to handle cases when `autofocus_motor_offset` is not None, improving the setup of autofocus devices.
- Refactor: Modified `axis_order` attribute in the `GeneratorMDASequence` class from a string to a tuple of strings, allowing for multiple axis orders to be specified.
- Test: Added new test function `test_set_autofocus_offset` to validate the functionality of the newly introduced `setAutoFocusOffset` method.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,7 +31,7 @@ repos:
       - id: mypy
         additional_dependencies:
           - pymmcore >=10.4.0.71.0
-          - useq-schema >= 0.4.0
+          - useq-schema >= 0.4.7
           - psygnal
           - typer
         files: "^src/"
--- a/src/pymmcore_plus/core/_mmcore_plus.py
+++ b/src/pymmcore_plus/core/_mmcore_plus.py
@@ -107,6 +107,29 @@ else:
 _OBJDEV_REGEX = re.compile("(.+)?(nosepiece|obj(ective)?)(turret)?s?", re.IGNORECASE)
 _CHANNEL_REGEX = re.compile("(chan{1,2}(el)?|filt(er)?)s?", re.IGNORECASE)
 
+# these are devices for which setAutoFocusOffset is known to have no effect
+# maps (device_library, device_name) -> offset_device_name
+# see _getAutoFocusOffsetDevice for details
+_OFFSET_DEVICES: dict[tuple[str, str], str] = {
+    ("NikonTE2000", "PerfectFocus"): "PFS-Offset",
+    ("NikonTI", "TIPFSStatus"): "TIPFSOffset",
+    # --------------------------
+    # these devices have an apparent offset device
+    # but may implicitly/privately control it just fine with the setOffset API
+    # need testing to see whether setAutoFocusOffset works as is.
+    # ("ZeissCAN29", "ZeissDefiniteFocus"): "ZeissDefiniteFocusOffset",
+    # ("Olympus", "AutoFocusZDC"): "ZDC2OffsetDrive",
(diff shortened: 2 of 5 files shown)


marktsuchida:
> @marktsuchida curious if you think this is a terrible idea :)

Not necessarily, but the real issue behind this is that there is no vendor-neutral concept of "offset" except in a very superficial sense. With some devices the offset varies while the Z stays at the same number; with others they change together (or the offset is in fact a Z value in micrometers). With some the Z readout is fixed during focus maintenance; with others the Z tracks the motor and changes. With some, you can save and restore offsets but you cannot compare them (they are not a 1D value). There are also differences in whether or not the offset can be queried and/or set while the focus maintenance is disabled.

So if you write code with one of these models, thinking that the "autofocus offset" means something specific, there is not much hope that your code will work  correctly on a different microscope. The devices are too different from each other, and the work of abstracting these differences in a way that would allow truly device-neutral code to manipulate offsets to be written simply hasn't been done yet.

As long as you are aware of this, there is nothing wrong with having a common get/set for the "offset" -- at least it will work if you do not compare or plot offsets or interpret them as any physical quantity, and only get/set them while the focus maintenance is enabled (I think, but I hesitate to even say this for sure...).

tlambert03:
> So if you write code with one of these models, thinking that the "autofocus offset" means something specific, there is not much hope that your code will work correctly on a different microscope.

yep, this i definitely get.  This isn't so much to make it so that the absolute value of autofocus offset means something more universal than it currently does.  It's more to make it that the core API call for applying some (vendor-specific) offset isn't *also* vendor specific (i.e. `setAutofocusOffset(...)` for some vendors and `setPosition(label, ...)` for other vendors).
Inasmuch as we can always guess what the correct call to `setPosition` *should* have been for an intended call to `setAutofocusOffset`, it seems harmless to redirect that call.

> at least it will work if you do not compare or plot offsets or interpret them as any physical quantity

👍

thanks!

tlambert03:
perhaps one non-obvious point of clarification.  The model that contains the offset will likely very often be generated by a GUI that is connected to the same microscope that runs the model.  So, at least in the case of pymmcore-plus & pymmcore-widgets.  The widget (which may or may not be connected to a core instance) can output a model with an offset value taken *from* the hardware into an intermediate object (the `MDASequence`) that  may then be consumed by that same hardware (in the engine).  So, in many cases, the relativeness of the offset is just fine

PR #276, opened 2023-10-21
feat: support DeviceInitializationState, fix issue with microscope model

tlambert03:
enum support for https://github.com/micro-manager/mmCoreAndDevices/pull/395


tlambert03:
this also fixes an issue with the microscope model tests resulting from https://github.com/micro-manager/mmCoreAndDevices/pull/376

see discussion in https://github.com/micro-manager/mmCoreAndDevices/issues/384

tlambert03:
the fix for widgets is in https://github.com/pymmcore-plus/pymmcore-widgets/pull/234

PR #287, opened 2023-11-16
fix: subclass apply_to_core method for PixelSizeGroup in model

fdrgsp:
At the moment, running PixelSizeGroup. apply_to_core(core) does not actually add any pixel size configuration to the core (and it does not give any error either). That's because in micromanager, defining a Group is different than defining a pixel configuration. 
This PR subclass the `ConfigGroup.apply_to_core` method of the `PixelSizeGroup` so that it can actually create new pixel size configuration.

### DIFF
--- a/src/pymmcore_plus/model/_pixel_size_config.py
+++ b/src/pymmcore_plus/model/_pixel_size_config.py
@@ -74,7 +74,8 @@ class PixelSizeGroup(ConfigGroup):
         for config_name, preset in self.presets.items():
             for dev, prop, val in preset.settings:
                 core.definePixelSizeConfig(config_name, dev, prop, val)
-                core.setPixelSizeUm(config_name, preset.pixel_size_um)
+            core.setPixelSizeUm(config_name, preset.pixel_size_um)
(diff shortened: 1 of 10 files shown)


src/pymmcore_plus/model/_pixel_size_config.py
```diff
@@ -59,3 +61,20 @@ def create_from_core(
...
+        *,
+        exclude: Container[str] = (),
+        on_err: ErrCallback | None = None,
+        apply_properties: bool = False,
+        then_update: bool = True,
+    ) -> None:
+        for config_name, preset in self.presets.items():
+            for dev, prop, val in preset.settings:
+                core.definePixelSizeConfig(config_name, dev, prop, val)
+                core.setPixelSizeUm(config_name, preset.pixel_size_um)
```
tlambert03, line 77: it looks like this line should probably not be inside the `for ... preset.settings` loop? 

also, it's missing `core.setPixelSizeAffine`

tlambert03 (changes requested):

tlambert03 (approved):

PR #293, opened 2023-12-02
fix: trying to fix windows tests

tlambert03:
something is happening with pyqt on windows.  trying to fix here


tlambert03:
fixed by moving to pyside, will investigate the pyqt thing elsewhere

PR #295, opened 2023-12-05
feat: Ensure Stage signals

wl-stepp:
Might be far from what you are thinking about. Will try on the scope this week and let you know if it works for me.


tlambert03:
thanks for getting this started @wl-stepp!
For the implementation, a slightly simpler pattern (and one similar to what we do with properties) might be to block any upstream signals, and always check (though maybe you explicitly decided against this sort of thing?):

```python
    @contextmanager
    def _stage_moved_emission_ensured(
        self, device: str | None = None, *args: Any
    ) -> Iterator[None]:
        """Context that emits events if any stage device moves."""
        device = device or self.getXYStageDevice()
        before = self.getXYPosition(device)
        with _blockSignal(self.events, self.events.XYStagePositionChanged):
            yield
            self.waitForDevice(device)

        after = self.getXYPosition(device)
        if before != after:
            self.events.XYStagePositionChanged.emit(device, *after)
```

However.  I think there's a important consideration here and I'm particularly curious to get @marktsuchida's thoughts on this one:
With stuff that takes time like a stage device, we have a difficult trade-off between enforcing blocking behavior on a call that was not previously blocking (i.e., using `waitForDevice` inside of `setXYPosition` has some unfortunate consequences) and missing actual change events.  @marktsuchida and I recently discussed that it might be possible to spin up a thread that polls the device in a thread and emits change events as it goes?  Of course, that implementation would be more complicated...

@marktsuchida what do you think here?  What is the "proper" implementation, and/or are there any "good enough" solutions that strike a good balance between implementation complexity and end-user usability (given what we've got to work with from the core and device adapters)

marktsuchida:
It's tough. When imagining doing this in the Core (C++), my thinking was that we would have a persistent background thread (or set of threads, possibly per-device-adapter). Every time we call a 'set' function (whether for a property or for stage position, the idea is the same), we would trigger a "wait-for-no-longer-busy" loop on the thread for that device before returning (without adding any blocking to the 'set' call itself).

Then, when the background thread detects that the device is not busy, it can 'get' the current value and issue a notification callback.

This can be made nicer by de-duplicating redundant notifications (such as those issued by devices, which may arrive before or after the core-generated one). Also, for things like motorized stage positions, we could periodically poll the in-motion position to provide more frequent notifications (but whether this is also a good idea for properties will need to be seen).

It is of course important to note that more than one 'set' can occur without necessarily waiting for the previous one to finish (at least with some devices) -- hence the persistent background thread, rather than one that is started every time a motion is initiated: we do not want many polling threads to be active at the same time.

I'm not sure there is a way to substantially simplify this without excessive blocking.

tlambert03:
thanks for your thoughts @marktsuchida.  Quick question: for playing around with this away from a scope. do any of the device adapters (demoCamera, sequenceTester, etc...) simulate a realistic (non-instantaneous) stage movement?

marktsuchida:
> for playing around with this away from a scope. do any of the device adapters (demoCamera, sequenceTester, etc...) simulate a realistic (non-instantaneous) stage movement?

DemoCamera does not have this, unfortunately. It does issue notifications, but synchronously (from within the `SetPositionUm` call). The demo XY stage does simulate a time delay proportional to the distance traveled (so you can "wait" for it) but the notification is synchronous so not a good representation of a real stage (and therefore I wouldn't count on this behavior not changing when it gets "fixed"). It also rejects attempts to start a new move while still busy, which might be how some stages behave.

SequenceTester never simulates time delay and never issues notifications, due to its purpose being the (automated) testing of acquisition engines (which cannot correctly rely on change notifications). However, it does record calls to `Busy()` so that you can troubleshoot missing waits for devices -- if you set something but fail to call `Busy()` (and get a `false` retval), this will be apparent in both the `STATE` and `HISTORY` sections (look for the `BUSY` entry of each device). (In fact, one of the ideas here is that you will always detect a forgotten call to `waitForDevice` because the device will stay "busy" forever in that case.)

But this is probably not that useful for testing synthetic notifications. What we perhaps want is a dedicated device adapter ("NotificationTester"?) that will simulate several different patterns of device behavior (based on what is attested in real device adapters), with or without notifications. I wouldn't mind writing this -- seems like a good tool to address a bunch of rough edges in MMStudio as well, not to mention a tool we'd need if we decide to synthesize notifications in the Core in the future. It would probably have the following configurable options:
- Whether or not the device becomes busy for some duration following the 'set' (as opposed to "completing the move" within the 'set' call)
- Whether or not change notifications are issued
- Whether change notifications are issued synchronously (on the thread calling the 'set' method) or asynchronously (on a device-owned thread)
- Whether the device issues the (final) change notification before or after transitioning to not-busy
- Possibly others

It would simulate real delays (or at least have the option to do so) but would not support hardware sequencing.

I know this sounds like a mess but it might be needed to dig ourselves out of the hole we're in :)

wl-stepp:
Thanks.
So on the scope I run into the problem that the live feed stops when I move the stage with this. I'm not 100% sure, but it seems like the camera snap misses the trigger and stays in snapImage(). Might not be a problem for internally triggered cameras...
I have to call setXYPosition without the ` @synchronized(_lock)` then it works. So might be a combination of waitForDevice and the lock. I hope I can investigate further later this week.

wl-stepp:
> For the implementation, a slightly simpler pattern (and one similar to what we do with properties) might be to block any upstream signals, and always check (though maybe you explicitly decided against this sort of thing?):

Well I thought it might be good to be the least intrusive possible. But true, it complicates things.
What about not even waiting for the device and just send something like a "XYStagePositionRequested". Then the user could decide to wait for the device or it's enough to know that the stage will have moved in a short time?

tlambert03:
hey @wl-stepp, happy new year and nice to see you hacking away at these things :)
We'd be happy to test some of this stuff out locally in the coming week or two as well.  Not sure whether our stages have adapters with similar event-emission implementations to what you're on (can you remind me what adapter you're on?)

> What about not even waiting for the device and just send something like a "XYStagePositionRequested". Then the user could decide to wait for the device or it's enough to know that the stage will have moved in a short time?

I think I like this idea.  It's sort of the the other bookend to the "wait-for-no-longer-busy" thread @marktsuchida was describing in https://github.com/pymmcore-plus/pymmcore-plus/pull/295#issuecomment-1841583959. It would be *much* easier to implement, and could act as a better-than-nothing approach until the (significantly harder) changes are made upstream.

Just to get my brain working again... can you describe how having that signal would solve your proximal issue that caused you to open this issue to begin with?

wl-stepp:
I came across the problem mainly here:
https://github.com/LEB-EPFL/iSIM/blob/48026a2e9ba89e8c5b2e88066d65600d2737108b/control/src/isim_control/gui/position_history.py#L85C9-L85C73

This makes an overview of the sample, not really crucial, but people have come to like it and I would like to keep it running. For this use case it's not important that we know exactly when the stage arrives.
In the stage widget I only get updates on the position with polling.
Positionhistory

marktsuchida:
A start of the NotificationTester I proposed above is now here: https://github.com/micro-manager/mmCoreAndDevices/pull/428 (though I don't have a stage simulator yet; that should be easy to add and I'll work on it next).

It should be noted that (1) get/set/busy ('busy' being what's behind `waitForDevice`) and (2) event notifications belong to two separate timelines, with no precise synchronization possible (other than that the first notification resulting from a 'set' will not happen before the 'set' -- obviously). This is just inherent in the distributed-system nature of the (device + PC) system.

When a stage capable of notifications moves, the last position notification may arrive before or after the device becomes non-busy. It is guaranteed that becoming non-busy _happens-after_ the device reaching the set position (or the move being canceled; otherwise it's a device adapter bug). It is also hopefully guaranteed that notifications _happen-after_ the device being observed at the notified position (I would consider it a bug for a device to send a notification for a target position that may not be reached yet -- although I wouldn't be surprised if we find such devices). But there is no further relation between the two events.

Also, devices may not guarantee to notify on small changes, so there is no general way to tell which notification (if there are any) is the last one resulting from a given 'set'. (It occurs to me that NotificationTester cannot simulate this aspect yet.) It gets even more nasty if notifications are significantly delayed such that the last one happens to arrive after the transition to non-busy: reading the position upon transition to non-busy may not be enough to obtain the quiescent position because a less accurate notification could arrive shortly later. (If we can know that the device will not move on its own (not true in general), we could decide to trust a get-when-not-busy more highly and not let a device-generated notification override it. In general, some polling is necessary for devices that can go quiescent without notifying the exact last position.)

In any case, I think it is best to stick to get/set/busy for control and sequencing purposes and use notifications (supplemented with polling when necessary) primarily for things like GUI display updates, avoiding trying to reconcile between the two if at all possible. This state of affairs is unlikely to change even if we make improvements to notifications in MMCore -- such improvements would only make GUI display updates work better without polling at the application level.

PR #297, opened 2023-12-08
docs: minor update

fdrgsp:
Minor docs update


fdrgsp:
> Why remove the cli stuff? Does it no longer work anymore?

I removed it because I tried to install just using `pip install pymmcore-plus` and I still can use the command line tool.  Can you try as well?

tlambert03:
ahh, looks like typer somehow got added to the main dependencies of `pymmcore-plus`: https://github.com/pymmcore-plus/pymmcore-plus/blob/47da10d277bbe01d1c52d2448fdea711fc28591d/pyproject.toml#L43C6-L43C11

but, I don't remember that being my intention.  So, please leave the docs here unchanged for now, and let's decide whether that was "accidental" or intentional before removing it here

docs/install.md
```diff
@@ -12,32 +12,30 @@ pip install pymmcore-plus
...
+    If you've installed with `pip install "pymmcore-plus[cli]"`, this library provides
+    a quick way to install the latest version of micro-manager:
 
     ```bash
     mmcore install
     ```
 
     This will download the latest release of micro-manager and place it in the
-    pymmcore-plus folder. If you would like to modify the location of the
+    `pymmcore-plus` folder. If you would like to modify the location of the
```
tlambert03, line 38: this actually isn't correct anymore... so while we're here, might as well update it to be correct.  explain that it goes into the user data folder (I think there's more info in the `install()` docstring)

docs/trableshooting.md
```diff
@@ -0,0 +1,24 @@
+# Trableshooting
```
tlambert03, line 1:
```suggestion
# Troubleshooting
```

docs/trableshooting.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Micro-Manager directory not found
+
+```sh
+pymmcore-plus - ERROR - (_util.py:131) could not find micromanager directory. Please run 'mmcore install'
+```
+
+If you tried to create a [`pymmcore_plus.CMMCorePlus`](api/cmmcoreplus.md) instance and got an
+error similar the one above, it probably means that you don't have Micro-Manager installed on your computer
```
tlambert03, line 10:
```suggestion
error similar the one above, it means that pymmcore-plus was unable to find micro-manager on on your system.
```

docs/trableshooting.md
```diff
@@ -0,0 +1,24 @@
...
+(for example, you can run `mmcore install` to install the latest version of Micro-Manager).
```
tlambert03, line 11:
```suggestion
Please run `mmcore install` to install the latest version of Micro-Manager.
```

mkdocs.yml
```diff
@@ -37,6 +37,7 @@ nav:
           - api/constants.md
           - Metadata: api/metadata.md
           - Utilities: api/utils.md
+      - trableshooting.md
```
tlambert03, line 40: fix typo: `troubleshooting.md` ... and fix file name

docs/install.md
fdrgsp: I've updated `install.md` with the new destination and I wanted to add an example path. I've used this
`C:\Users\UserName\AppData\Local\pymmcore-plus\mm` but I'm not sure if it is correct, I need to check in a windows pc.

PR #299, opened 2023-12-12
feat: update OMEZarrWriter to position-specific sub-MDAsequences

fdrgsp:
@tlambert03 in this PR I've added the ability to use the `OMEZarrWriter` also with `MDASequences` that have positions with `sub-MDASequences`. When you'll have some time, let me know what do you think about this approach.

I also updated the tests.

### DIFF
--- a/src/pymmcore_plus/mda/handlers/_ome_zarr_writer.py
+++ b/src/pymmcore_plus/mda/handlers/_ome_zarr_writer.py
@@ -219,7 +219,7 @@ class OMEZarrWriter:
 
     def sequenceStarted(self, seq: useq.MDASequence) -> None:
         """On sequence started, simply store the sequence."""
-        self._current_sequence = seq
+        self._set_sequence(seq)
 
     def sequenceFinished(self, seq: useq.MDASequence) -> None:
         """On sequence finished, clear the current sequence."""
@@ -234,7 +234,6 @@ class OMEZarrWriter:
         """Write frame to the zarr array for the appropriate position."""
         # get the position key to store the array in the group
         key = f'{POS_PREFIX}{event.index.get("p", 0)}'
-
         if key in self._arrays:
             ary = self._arrays[key]
         else:
@@ -242,11 +241,11 @@ class OMEZarrWriter:
             # create a new Zarr array in the group for it
             if not self._current_sequence:
                 # just in case sequenceStarted wasn't called
-                self._current_sequence = event.sequence
+                self._set_sequence(event.sequence)
 
             if not (seq := self._current_sequence):
                 # This needs to be implemented for cases where we're not executing
-                # an MDASequence. Or, we need to create a better "mock" MDASequence
+                # an MDASequence.  Or, we need to create a better "mock" MDASequence
                 # for generic Iterable[MDAEvent]
                 raise NotImplementedError(
                     "Writing zarr without a MDASequence not yet implemented"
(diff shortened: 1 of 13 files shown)


tlambert03:
thanks for this @fdrgsp!  I haven't yet looked too closely at the implementation, I just refactored the tests a bit to update the parameterization and make the link between the input data and the expectation a bit clearer and decoupled from the test itself.  Can you have a quick look at https://github.com/pymmcore-plus/pymmcore-plus/pull/299/commits/ed593e3eb88f33eb4017ea684deec3a859ba5127 and make sure that the expectations looks like you expect?

tlambert03 (commented):
i think one thought I'm having as I look at this is that, while this may have come to your attention in the Zarr writer, the thing you're trying to solve here is really much more general, and could possibly live in useq-schema itself.  The issue is that `MDASequence.sizes` isn't correct for plans with subsequences (an issue could be opened at useq about that).  Now, that issue is kinda hard to fix globally without being backwards incompatible, since a single dict of `{axis -> size}` will no longer work.  But, we can make an additional utility function or method like:

```python
def position_sizes(seq: MDASequence) -> list[dict]:
    """Return a list of size dicts for each position in the sequence."""
    sizes = []
    main_sizes = seq.sizes
    main_sizes.pop("p", None)  # remove position
    for p in seq.stage_positions:
        if p.sequence is not None:
            psizes = {k: v or main_sizes.get(k, 0) for k, v in p.sequence.sizes.items()}
        else:
            psizes = main_sizes.copy()
        sizes.append({k: v for k, v in psizes.items() if v})
    return sizes
```
 
which, when used:

```python
seq = MDASequence(
    axis_order="tpcz",
    channels=["Cy5", "FITC"],
    stage_positions=[
        (1, 2, 3),
        {"x": 4, "y": 5, "sequence": MDASequence(grid_plan={"rows": 2, "columns": 1})},
    ],
    z_plan={"range": 3, "step": 1},
)

print(position_sizes(seq))
# [{'c': 2, 'z': 4}, {'g': 2, 'c': 2, 'z': 4}]
```

that would have the benefits of 
A) pulling most of the logic out of the Zarr writer here 
B) being easier to test (since you can test that you're correctly inferring sub-position sizes without running the full zarr writer)
C) probably reducing a fair amount of code here.

can you try something like that function as a module level function within in this PR as a start?

fdrgsp:
> can you try something like that function as a module level function within in this PR as a start? 

👍🏻 yes I'll try to do that!

fdrgsp:
> C) probably reducing a fair amount of code here.

@tlambert03 indeed much shorter :)

src/pymmcore_plus/mda/handlers/_ome_zarr_writer.py
```diff
@@ -208,40 +232,50 @@ def frameReady(
...
+            ary = self._new_array(key, shape, frame.dtype, used_axes)
+
+            # write the MDASequence metadata and xarray _ARRAY_DIMENSIONS to the array.
+            # _ARRAY_DIMENSIONS will be used to get the correct index order when writing
+            # to disk.
             ary.attrs.update(
                 {
                     "useq_MDASequence": json.loads(seq.json(exclude_unset=True)),
-                    "_ARRAY_DIMENSIONS": axes,
+                    "_ARRAY_DIMENSIONS": used_axes,
```
tlambert03, line 272: this is incorrect.  `_ARRAY_DIMENSIONS` needs to have same length as shape (meaning it needs to include "y" and "x".   This key is part of the xarray zarr spec: https://docs.xarray.dev/en/latest/internals/zarr-encoding-spec.html ... and this change breaks that.

Looks like that wasn't explicitly tested before, but this needs to be changed back to how it used to be (and the tests need to be changed to make sure that `_ARRAY_DIMENSIONS` includes `y` and `x`)... and that might mean you also need to undo the change below that uses `_ARRAY_DIMENSIONS` to get the index

tlambert03:
i updated the test expectations here (which will break the tests) to reflect what the `_ARRAY_DIMENSIONS` should actually be

tlambert03 (approved):
made a couple small changes to reduce the diff size further, and updated tests slightly.  Thanks @fdrgsp!

PR #308, opened 2024-01-24
fix: fix available_peripherals for loaded hub devices

tlambert03:
This is a partial fix to a bug in the pymmcore-widgets ConfigWizard that results in some hub devices (like SequenceTester) not showing any peripherals.

The problem is that some hub devices reveal their peripherals as top level library devices, while others may wait until after device initialization to reveal them.

What this PR does is to make sure that calling `Microscope.load_available_devices()` will not only check `getAvailableDevices` for all `getDeviceAdapterNames` ... but will *also* check `getInstalledDevices` for all `getLoadedDevicesOfType(DeviceType.Hub)`.  

it's not a complete for for widgets though, since the ConfigWizard will still need to call `model.load_available_devices()` after successfully initializing a hub device (which will happen over in that repo)

### DIFF
--- a/src/pymmcore_plus/_util.py
+++ b/src/pymmcore_plus/_util.py
@@ -142,7 +142,12 @@ def find_micromanager(return_first: bool = True) -> str | None | list[str]:
 def _qt_app_is_running() -> bool:
     for modname in {"PyQt5", "PySide2", "PyQt6", "PySide6"}:
         if modname in sys.modules:
-            QtWidgets = importlib.import_module(".QtWidgets", modname)
+            try:
+                # in broken environments modname can be a namespace package...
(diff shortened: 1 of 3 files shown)


marktsuchida:
> some hub devices reveal their peripherals as top level library devices, while others may wait until after device initialization to reveal them

This is indeed confusing. The canonical behavior is meant to be that they are only available after initialization and through `getInstalledDevices()`. The reason some show up as top-level devices is because those device adapters predate the hub-peripheral interface (or copied other device adapters that do). (We should document this better, of course.)

tlambert03:
yep, I figured it was an evolution of an API.  Makes sense, and not a big problem for me!

PR #309, opened 2024-01-25
fix: fix available peripherals for libraries with multiple hubs

tlambert03:
fixes a silly bug in my code that fails to link available peripherals with their parent hub device when the library has multiple hubs.  Test will only work locally for now as it uses the Ti2 device adapter, but we can probably find another example of a multi-hub library that's easy to test.


marktsuchida:
It looks like `BH_DCC_DCU` is the only other one, based on a crude grep. It has 2 fixed hubs that can be created without hardware. I don't remember if it initializes without any hardware connected -- it might, in which case it will detect no peripherals (so perhaps not great for testing). It also requires a vendor DLL and is Windows-only.

I saw this and was surprised to learn that we support having more than one hub device in a single device adapter, and then realized that both examples were my doing....

PR #310, opened 2024-01-26
feat: add SequenceTester to apple silicon build-dev

tlambert03:
adds SequenceTester to the list of packages built when running `mmcore build-dev` on apple silicon


tlambert03:
currently builds fine but doesn't work on load. `RuntimeError: Failed to load device adapter "SequenceTester"` ... not sure why yet

marktsuchida:
> `RuntimeError: Failed to load device adapter "SequenceTester"`

I'm pretty sure the original exception string has the root cause appended. Any chance you can see something in the CoreLog?

PR #312, opened 2024-01-28
refactor!: more minimal, and overrideable MDA metadata

tlambert03:
Metadata handling in the MDA engine is still a bit haphazard.  We largely copied the state of things in MM-studio, and discussions with multiple MM core devs has suggested that perhaps that isn't necessarily the best approach.  MM grabs a *lot* of metadata for every frame, much of which doesn't change over the course of the experiment, and it also will almost always match the metadata in the `MDAEvent` "command" that triggered that given frame.   Nico has suggested that this comes at a pretty big penalty, especially for very fast acquisition.

While this PR doesn't really formalize any approach just yet it does two things:
1) establishes two methods in the MDAEngine that are more easy for a subclass to override and customize:
    - `get_summary_metadata()` - called once at the beginning
    - `get_frame_metadata()` - called for every frame
2) strips some (but not all) of what was being called for `get_frame_metadata`.  Notably `getSystemStateCache` is still there, which is probably unnecessary, but which can be discussed in a follow up.

In general, I would very much like to have a better schema for the format (including keys) of the metadata... but that's a bigger issue, and this PR just simplifies that creation of the metadata.

@wl-stepp, I'm curious to what extent you've begun "depending" on a specific format of information in that third parameter to the `frameReady` event?

### DIFF
--- a/src/pymmcore_plus/core/_mmcore_plus.py
+++ b/src/pymmcore_plus/core/_mmcore_plus.py
@@ -1536,7 +1536,7 @@ class CMMCorePlus(pymmcore.CMMCore):
         # you shouldn't have to search the code to find out what keys are available
 
         tags = dict(meta) if meta else {}
-        for dev, label, val in self.getSystemStateCache():  # type: ignore
+        for dev, label, val in self.getSystemStateCache():
             tags[f"{dev}-{label}"] = val
 
         tags["BitDepth"] = self.getImageBitDepth()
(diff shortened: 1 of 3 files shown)


wl-stepp:
Not too much for now, no worries. Mostly for getting 'ElapsedTime-ms'. But I can test most of the things off the scope, so should be ok to adjust. I like the idea of a more lean approach here.

tlambert03:
going to merge this.  This is a potentially breaking change for anyone already depending on the metadata structure, but I'm hoping that's almost no one at this point, and better to start minimal and add later than remove later

PR #313, opened 2024-01-28
feat: Add file saving `output` argument to MDARunner.run

### LINKED ISSUE #311: Request for opinions on data-saving APIs
This is a call for opinions.  
CC @fdrgsp @ianhi @wl-stepp @kmdouglass @marktsuchida @dpshepherd @nng-thienphu @linshaova
(it's ok if you don't feel that up-to-date on the goings on of pymmcore-plus, it's kind of a high-level question)

With a protocol for data handlers/writers (like `OMEZarrWriter` and `ImageSequenceWriter`) in place, we need a higher level API for indicating to the runner how it should save data.  Currently, the MDA runner just "emits" frames as they are acquired and it's up to the end-user to attach any specific data handlers they want; this would aim to add a convenience layer on top of that.

## API for Specifying a Writer

There is already the low-level:

```python
writer = OMEZarrWriter("file.zarr")
with mda_listeners_connected(writer):
    core.mda.run(sequence)
``` 

but I was always hoping to offer a slightly nicer API than that.  Here are a few options (and note that all of these are mutually compatible with a user *also* attaching additional listeners using `mda_listeners_connected`)

### 1. Add an `output` field to `MDARunner.run`

(can also be named something other than `output`)

```python
core = CMMCorePlus()
core.mda.run(..., output='some/path.ome.zarr')  # pass a string
# or 
my_writer = CustomDataHandler()
core.mda.run(..., output= my_writer). # pass a data handler object 
```

The argument would be either:
- `str` - in which case a builtin Handler would be used based on extension.  Note that fsspec can be used to support remote filesystems.
    - `ome.zarr` - > `OMEZarrWriter`
    - `ome.tiff` -> `OMETiffWriter`
    - `dir` -> no extension would use `ImageSequenceWriter`
- A Handler object, (anything implementing the data handler protocol, meaning it has at least a `frameReady` method that accepts 1-3 arguments).  This could be, for example, something that would deskew and rotate images prior to writing them for @dpshepherd

### 2. Add an `output` field to `useq.MDASequence`

A similar thing could be added at a higher level to useq-schema's `MDASequence` object

```python
core = CMMCorePlus()
seq = MDASequence(output='some/path.ome.zarr', ...)
core.mda.run(seq)
```

The type semantics of `str` would be the same.  But since `useq-schema` should remain agnostic to the APIs of `pymmcore-plus`, it's harder to say exactly how a custom writer should be declared there in a way that retains `useq-schemas` agnosticism to micro-manager, imswitch, etc...

Also, as it requires an `MDASequence` object (which might not always be created in the cases of event-driven or non-deterministic sequences like what @wl-stepp and @kmdouglass might use) it might need to be done *in addition* to explicitly allowing `output` to be passed to the `run()` method.  That is, `run()` will first look to see if an `output` argument has been passed, and then fallback to an `MDASequence.output` field if it has been passed.

### 3. Give `MDARunner` a new `writer` or `handler` attribute that will always be used:

```python
core = CMMCorePlus()
my_writer = CustomDataHandler()
core.mda.set_handler(my_writer). # or `set_writer` or something
core.mda.run(...)
```

Note that this is rather similar to just using:

```python
my_writer = CustomDataHandler()
with mda_listeners_connected(writer):
    core.mda.run(...)
```

except that it permanently attaches `my_writer` to the runner, and you could call `run()` multiple times.
It doesn't prevent a user from attaching additional handlers if desired, it just essentially acknowledges that there is a usually a "special" handler that will be responsible for writing.

## File name incrementing

In all cases, we have to determine how the file name will be incremented if called twice, similar to @ianhi's `get_unique_folder` method [here](https://github.com/pymmcore-plus/pymmcore-MDA-writers/blob/e46d3ba0fc777393445b1b576a0f592e76fc067e/pymmcore_mda_writers/_writers.py#L48).  Any ideas welcome here.

1. It could be left entirely to the Handler: each handler would be responsible for checking if the destination exists, and changing the name or raising accordingly.  This gives the most flexibility, but makes writing a handler significantly more cumbersome.
2. Add a new argument somewhere, perhaps `run(..., output='some_path', increment=some_function)`: where `some_function` is a `Callable[[str], str]]` - a function that takes the requested string and returns a new useable string (or raises).  This could have a default implementation that adds `_001` for example to the filename.
3. Just have pymmcore-plus magically increment filenames that already exist.

tlambert03:
This implements option number 1 and closes #311 

### 1. Add an `output` field to `MDARunner.run`

```python
core = CMMCorePlus()
core.mda.run(..., output='some/path.ome.zarr')  # pass a string
# or 
my_writer = CustomDataHandler()
core.mda.run(..., output= my_writer). # pass a data handler object 
```

The argument would be either:
- `str` - in which case a builtin Handler would be used based on extension.  Note that fsspec can be used to support remote filesystems.
    - `ome.zarr` - > `OMEZarrWriter`
    - `ome.tiff` -> `OMETiffWriter`
    - `dir` -> no extension would use `ImageSequenceWriter`
- A Handler object, (anything implementing the data handler protocol, meaning it has at least a `frameReady` method that accepts 1-3 arguments).  This could be, for example, something that would deskew and rotate images prior to writing them.

### DIFF
--- a/src/pymmcore_plus/core/_mmcore_plus.py
+++ b/src/pymmcore_plus/core/_mmcore_plus.py
@@ -54,6 +54,8 @@ if TYPE_CHECKING:
     from typing_extensions import Literal, TypedDict
     from useq import MDAEvent
 
+    from pymmcore_plus.mda._runner import SingleOutput
+
     from ._state import StateDict
 
     _T = TypeVar("_T")
@@ -1411,7 +1413,11 @@ class CMMCorePlus(pymmcore.CMMCore):
         return self._mda_runner
 
     def run_mda(
-        self, events: Iterable[MDAEvent], *, output=None, block: bool = False
+        self,
+        events: Iterable[MDAEvent],
+        *,
+        output: SingleOutput | Sequence[SingleOutput] | None = None,
+        block: bool = False,
     ) -> Thread:
         """Run a sequence of [useq.MDAEvent][] on a new thread.
 
@@ -1428,6 +1434,16 @@ class CMMCorePlus(pymmcore.CMMCore):
         events : Iterable[useq.MDAEvent]
             An iterable of [useq.MDAEvent][] to execute.  This may be an instance
             of [useq.MDASequence][], or any other iterable of [useq.MDAEvent][].
+        output : SingleOutput | Sequence[SingleOutput] | None, optional
+            The output handler(s) to use.  If None, no output will be saved.
+            "SingleOutput" can be any of the following:
+
+            - A string or Path to a directory to save images to. A handler will be
+                created automatically based on the extension of the path.
+            - A handler object that implements the `DataHandler` protocol, currently
+                meaning it has a `frameReady` method.  See `mda_listeners_connected`
+                for more details.
+            - A sequence of either of the above. (all will be connected)
         block : bool, optional
             If True, block until the sequence is complete, by default False.
 
--- a/src/pymmcore_plus/core/_mmcore_plus.py
+++ b/src/pymmcore_plus/core/_mmcore_plus.py
@@ -1467,7 +1467,7 @@ class CMMCorePlus(pymmcore.CMMCore):
             raise ValueError(
                 "Cannot start an MDA while the previous MDA is still running."
             )
-        th = Thread(target=self.mda.run, args=(events, output))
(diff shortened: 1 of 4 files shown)


tlambert03:
this has been open long enough, going to merge.  hope it works :joy:

PR #314, opened 2024-01-28
refactor: move imageSnapped event

### LINKED ISSUE #144: should `events.imageSnapped` be emitted on `snapImage()` rather than `snap()`
currently, the `imageSnapped` event _only_ gets emitted if someone use the `snap()` method that is introduced by pymmcore-plus ... but not if they use the `snapImage()` method that they're probably accustomed to using.  This makes it less useful for saying "an image is ready to be viewed", as it now depends very much on how the end-user is using pymmcore-plus.

I think we should perhaps change the `imageSnapped` event to be emitted by the `snapImage()` method, and have it emit nothing, rather than only upon using our custom `snap()` method.

then it would be:
```
@core.events.imageSnapped.connect
def _handle():
    core.getImage()
```

thoughts @ianhi @fdrgsp ?

tlambert03:
this closes #144 by moving the `imageSnapped()` event from the `CMMCorePlus.snap()` method (which is new and unexpected for most people) to the `snapImage()` method (which is both more familiar and semantically correct).

This change is partially motivated by the confusion caused by this in https://github.com/pymmcore-plus/napari-micromanager/issues/305

Some considerations before merge:
1. since `snapImage()` no longer has access to the resulting image, it removes the numpy array as an argument to the callback, which is a breaking change.  Callbacks must now call `core.getImage()` themselves to retrieve the waiting image.  A [github code search](https://github.com/search?type=code&auto_enroll=true&q=imageSnapped.connect) result suggests we're the only ones using it, so not a big deal... but @wl-stepp has [one line of commented-out code](https://github.com/LEB-EPFL/iSIM/blob/856a2d67eecb27b7e8680e13c7086e55d8a3e9ff/control/src/isim_control/gui/position_history.py#L126) that would break if the comment were removed
2. We need to double check that this doesn't play poorly with an in-process MDA, which will also call snapImage periodically.  It's slightly risky in that regard, since someone could carelessly connect something to snapImage and accidentally slow down or otherwise hinder their experiment?  Probably unlikely, but worth thinking about.

### DIFF
--- a/src/pymmcore_plus/core/_mmcore_plus.py
+++ b/src/pymmcore_plus/core/_mmcore_plus.py
@@ -1391,11 +1391,11 @@ class CMMCorePlus(pymmcore.CMMCore):
 
         **Why Override?** To add a lock to prevent concurrent calls across threads.
         """
-        autoshutter = self.getAutoShutter()
-        if autoshutter:
+        if autoshutter := self.getAutoShutter():
             self.events.propertyChanged.emit(self.getShutterDevice(), "State", True)
         try:
             super().snapImage()
+            self.events.imageSnapped.emit()
         finally:
             if autoshutter:
                 self.events.propertyChanged.emit(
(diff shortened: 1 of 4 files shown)


tlambert03:
as expected widgets tests are failing with this change,  will make a PR to pymmcore-widgets that will be compatible with both callback signatures (by having it take no arguments and call `getImage()` itself)

wl-stepp:
Thanks for the heads up, no problem for me.

PR #315, opened 2024-01-31
fix: only allow 'AcquireImage' Actions (or None) to be sequenced

fdrgsp:
This is only relevant when running a `SequencedEvent`  (use_hardware_sequencing=True). 

This PR updates the `can_sequence_events` method with `useq` `Actions`. Only instances of `AcquireImage` `Action` (or None) can be sequenced.


fdrgsp:
Also in the `exec_sequenced_event` method , shouldn't we check if the acquisition has been paused or cancelled? 

https://github.com/pymmcore-plus/pymmcore-plus/blob/edda9cbc3c241957473bde4d67ad1a87a834a0ed/src/pymmcore_plus/mda/_engine.py#L325-L372

something like :

```py
...
self._mmc.startSequenceAcquisition(
            n_events,
            0,  # intervalMS  # TODO: add support for this
            True,  # stopOnOverflow
        )

 count = 0
        iter_events = iter(event.events)
        # block until the sequence is done, popping images in the meantime
        while self._mmc.isSequenceRunning():
            if self._mmc.getRemainingImageCount():
                img = self._mmc.popNextTaggedImage()
                e = next(iter_events)
                yield ImagePayload(img.pix, e, img.tags)
                count += 1

                if self._mmc.mda._wait_until_event(  # noqa
                    event.events[int(img.tags["ImageNumber"])]
                ):
                    self._mmc.mda.cancel()
                    self._mmc.stopSequenceAcquisition()

            else:
                time.sleep(0.001)
...
```

src/pymmcore_plus/mda/_engine.py
```diff
@@ -149,6 +149,14 @@ def exec_event(self, event: MDAEvent) -> Iterable[PImagePayload]:
                 # store correction for this position index
                 p_idx = event.index.get("p", None)
                 self._z_correction[p_idx] = new_correction
+
+            # this is necessary for SequencedEvent, if we don't yield
+            # from self.exec_sequenced_event(event) the acquisition stops
+            # because we return ()
+            if isinstance(event, SequencedEvent):
```
tlambert03, line 156: at first glance, i find it a bit odd to be checking `if isinstance(event, SequencedEvent):` inside of this `HardwareAutofocus` conditional, given that we check it again immediately below, and perform the same action `yield from self.exec_sequenced_event(event)`.  So it kinda feels to me like the `if` blocks themselves need evaluation.

One question I have is: "what is the meaning of a `SequencedEvent` that has a `HardwareAutofocus)` action?".  That is, why do we ever get here in the first place?  It's possible that we need to re-examine the building of a `SequenceEvent` and special case other actions...  Where you able to see more about the event that caused this?  Was it in fact a multi-frame event that also had a hardware action?

src/pymmcore_plus/core/_sequencing.py
```diff
@@ -228,6 +228,12 @@ def can_sequence_events(
     def _nope(reason: str) -> tuple[bool, str] | bool:
         return (False, reason) if return_reason else False
 
+    # Action
+    if (e1.action is None or not isinstance(e1.action, AcquireImage)) or (
+        e2.action is None or not isinstance(e2.action, AcquireImage)
+    ):
```
tlambert03, line 234:
```suggestion
    # Action
    if not isinstance(e1.action, (AcquireImage, type(None))) or not isinstance(
        e2.action, (AcquireImage, type(None))
    ):
```

this doesn't feel quite right.  I thought the goal was to *allow* `None`, not to forbid it.

tlambert03:
can you have a look at the comment, and then update the title of the PR to reflect the changed content of the PR?

PR #326, opened 2024-02-13
fix: re-engage hardware autofocus after performing autofocus action if it was engaged

fdrgsp:
If the hardware autofocus is engaged when starting the acquisition, we should keep it engaged after running the autofocus action.

For example, if we want to do a 100 frames fast timelapse (0ms interval) and we want to use the autofocus, then the autofocus should be on for the entire acquisition to avoid axial shift.


tlambert03:
looks good thanks,  made a small change: since you were always only updating one of the two items in the tuple (and using a private `_replace` method to do so), I think it's clearer to just make two variables: `self._af_was_engaged` and `self._af_succeeded`

Issue #87, opened 2022-02-26
make `pymmcore_plus.install` work for linux


tlambert03:
@ianhi, this isn't something I've done before... how do you install for linux?

ianhi:
also you can't get the secret device adapters, only the private ones :(

ianhi:
part of why I want https://github.com/micro-manager/mmCoreAndDevices/issues/87

tlambert03:
so you did!
gotcha.  we've got some amazing people building all these conda recipes for napari plugins these days... maybe I can bug them to do this

ianhi:
It's a bit tricky because the build system there is pretty tightly coupled to the micro-manager build system. You can see about that here: https://github.com/micro-manager/mmCoreAndDevices/pull/86 and in particualr: https://github.com/micro-manager/mmCoreAndDevices/pull/86#issuecomment-1033224144 and https://github.com/micro-manager/micro-manager/issues/1392#issuecomment-1033140970

so the sad reality is probably that until those changes get made it's not oging to be easy to do on linux :(

ianhi:
Although for CI at least all we really need is the DemoCamera `.so` file which for binder at least has been done and you can distribute like so: https://github.com/imjoy-team/micro-manager-imjoy/blob/42f982c8f360a6ba6a0c9f3d453ec81c2a7e48a5/postBuild

So we could compile that once and re-use in CI

tlambert03:
thanks for your thoughts!

> On Linux, there is no single rule for "what the operating system provides", because that depends on the distribution (and what packages have been installed by the user).
> Conda presumably solves these issues by building its own world outside of the distro package manager (which is good and bad, depending on whether you live in that world or not).

yep. this is definitely one advantage of distributing with conda.  because it also distributes standard (non-python) libraries, one can declare all of the dependencies, test them in a minimal docker container, and make sure that they will come along with all of the needed C libraries, while avoiding binary incompatibilities.  I'd never advocate that you _force_ people to install with conda (which is the only bad part i can see with that scenario)... but there is definitely a huge part of the scientific community who might want to use mmcore that are "in" that world.  So once we get it working, it could be as simple as `conda install napari-micromanager` and you've got a full, platform-independent, python/C gui for micromanager.

anyway, I do see that you'll need to continue to develop/maintain all of these alternate approaches for maximum flexibility.  But I might also try to tag some folks to help us get it all building on conda.  For those in that world, it would definitely be convenient.

[...]

PR #331, opened 2024-03-01
feat: support linux on `mmcore build-dev`, test on linux, make device list editable

tlambert03:
related to #87

while `mmcore install` won't work for linux, `mmcore build-dev` will.  But, it's not the "full package" with all the device adapters, so I'm not marking it as solved yet

closes #310  because it also allows the device list to be specified at build time

**update**:
while the original intent of this PR was just to get device adapters building on linux, it has expanded into a PR that makes a lot of general improvements to `mmcore build-dev`
- works for mac and linux now
- accepts any list of device adapters: `mmcore build-dev DemoCamera SequenceTester ...`
- will validate those device adapter names early in the process
- better fetching of git sha for early detection of existing destination directory
- better modification of the configure.ac and makefile.am files (no more local duplication of those files, just edits them dynamically)

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -60,7 +60,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ~/.local/share/pymmcore-plus/
-          key: ${{ runner.os }}-mmbuild-${{ hashFiles('src/pymmcore-plus/_build.py') }}
+          key: ${{ runner.os }}-mmbuild-${{ hashFiles('src/pymmcore_plus/_build.py') }}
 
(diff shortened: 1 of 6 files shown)


tlambert03:
looks like everything's green here!  Hooray, regular testing on linux here too.  This will also make benchmarks easier (#330) because i think codspeed is only supporting ubuntu runners at the moment

tlambert03:
@marktsuchida, just a friendly heads up here that this PR makes `mmcore build-dev` significantly more useful, and quicker, for building a subset of device adapters on any mac/linux computer, and adds `SequenceTester` (along with DemoCamera and Utilities) to the default list

PR #339, opened 2024-03-07
feat: [WIP] PyDevice concept and experimenting with acquire-imaging

tlambert03:
hey hey @nclack! 
(and @marktsuchida you might also be interested, but please don't be too worried by my explorations! :joy:)

here's my very very WIP stab on thinking about how I would integrate acquire-python into pymmcore-plus.

There's two things broadly going on here:

1. I'm fumbling my way around an `AcquireCamera` class that represents a runtime representation of an acquire camera that I can drive from the [`MDAEngine`](https://github.com/pymmcore-plus/pymmcore-plus/blob/fedd29c0c89307e842fefa72fcb8b58b8a884699/src/pymmcore_plus/mda/_engine.py#L54).  Mostly, i (think) I need it to be able to setup various device parameters, snap a single image (I know this is the part that is least compatible with your design), and setup/execute a streaming acquisition, where the camera is either a master or slave expecting a certain number of triggers. @nclack, it would be awesome if you could have a quick look at the `acquire_cam.py` file here and tell me if I'm doing anything that makes you want to 🤮
2. The other part concerns acquire a bit less (and will certainly concern @marktsuchida a bit more :joy:). It is playing with how we could integrate pure python (non-MM) devices into pymmcore-plus.  For that I'm:
   - overloading the `loadSystemConfiguration` method to add new "commands" to the config file specification.  Specifically, something like `PyDevice,Camera,pymmcore_plus.devices.acquire_cam,AcquireCamera` would mean "load the `AcquireCamera` class from the `pymmcore_plus.devices.acquire_cam` module (interface to be determined).  I'm playing with the idea that a single config could be used for both micro-manager Java, and pymmcore-plus (i.e. we'd be a super-set that would do stuff on top of the existing spec, but should fallback gracefully... so the command might be `#PyDevice` instead)
   - overloading various CMMCorePlus methods (the one that take a device label as the first argument) to first check whether the device is one of our own python-devices, and if so, use it's API instead of the MMCore API.  This is where it gets tricky, but for the moment, if we're simply restricting ourselves to cameras and the pymmcore-plus MDAEngine, it's definitely possible that we could get it to be "good enough" to incorporate non MMcore camera devices (probably the highest value target)


tlambert03:
superceded by #407

PR #343, opened 2024-04-24
[WIP] feat: add support for Micro-Manager Multi Camera Utilities

fdrgsp:
related to https://github.com/pymmcore-plus/napari-micromanager/pull/333


src/pymmcore_plus/mda/_engine.py
```diff
@@ -285,10 +285,14 @@ def exec_single_event(self, event: MDAEvent) -> Iterator[PImagePayload]:
             return ()
         if not event.keep_shutter_open:
             self._mmc.setShutterOpen(False)
-        yield ImagePayload(self._mmc.getImage(), event, self.get_frame_metadata())
+
+        for cam in range(self._mmc.getNumberOfCameraChannels()):
+            data, meta = self._mmc.getTaggedImage(cam)
```
tlambert03, line 290: i'd really like to avoid using `getTaggedImage` internally if possible.  It was added to mimic the java stuff, but it's very slow and exposes a ton of keys that are not really well defined (so I don't want to encourage people to use them yet).

Can you try to look deeper into the code, perhaps in `popNextImageAndMD` or in `getTags` and see if you can add the bare minimum needed to get_frame_metadata here?

I suspect it's these lines:
https://github.com/pymmcore-plus/pymmcore-plus/blob/e77d8301485daf9650a51cc6333f3b069e96cdf6/src/pymmcore_plus/core/_mmcore_plus.py#L1576-L1580

wl-stepp:
Are there any updates on this? Might become relevant for us in July.

fdrgsp:
> Are there any updates on this? Might become relevant for us in July.

Hi Willi, I'll keep working on it then! Are you planning to use in with hardware sequencing?

wl-stepp:
No, no hardware sequencing I think. Let me know if there is something I can help with.

fdrgsp:
NOTE: I had a look and I think that #351 is also implementing all these changes made here (because of the new metadata handling), so I guess once that one is ready, we can close this.

tlambert03:
yeah, it's all tied up together in there :)  
that PR is shaping up... the biggest issue I'm struggling with at the moment is whether or not to change the actual object type from a dict to a struct.  dicts have the advantage of being standard library and general purpose, but the struct is nicer for many reasons including clarity, serialization, and speed.  Moving to a struct kind of breaks all of our current handlers, and moreover, all handlers kinda need to know up front what kind of metadata they're going to be dealing with... so, it's an ecosystem compatibility challenge at the moment

tlambert03:
this should be superceded now by #358

PR #348, opened 2024-05-10
feat: add tensorstore writer

tlambert03:
This writer/handler is similar to (and I suspect will eventually replace) the `OMEZarrWriter`.  It uses tensorstore to write to disk, which in my quick-and-dirty benchmarks seems to be writing about 3x faster than python-zarr (note, i haven't tried to tune the python-zarr much to make it faster, but tensorstore is written in C++ and is explicitly designed to take care of multi-threading, and just generally be performant).  

Tensorstore supports a number of formats, though zarr is what we'll likely use the most. I'm not currently worrying about writing out to the ngff spec, as this writer is designed to handle more than 5 dimensions easily (since tensorstore also supports that nicely) and I don't want to artificially restrict it yet... we can add an ome-zarr variant in the future.

This writer has two main "modes" (which are determined implicitly based on the experiment).  When `self._nd_storage == True`, we're greedily attempting to store the data in a multi-dimensional format based on the axes of the sequence.  For non-deterministic  event-driven experiments, this often won't work... so `self._nd_storage == False` means it will simply store frames as they come, with a final shape of ` (nframes, y, x)`.  (It should then be relatively straightforward to reshape the data after the fact if need be).  

This writer was partially created while developing the stack-viewer in pymmcore-widgets: https://github.com/pymmcore-plus/pymmcore-widgets/pull/293 and will be used as the default data store for that.

### DIFF
--- a/examples/tensorstore_mda.py
+++ b/examples/tensorstore_mda.py
@@ -10,13 +10,13 @@ core.loadSystemConfiguration()
 
 sequence = MDASequence(
     channels=["DAPI", {"config": "FITC", "exposure": 1}],
-    # stage_positions=[{"x": 1, "y": 1, "name": "some position"}, {"x": 0, "y": 0}],
+    stage_positions=[{"x": 1, "y": 1, "name": "some position"}, {"x": 0, "y": 0}],
     time_plan={"interval": 2, "loops": 3},
     z_plan={"range": 4, "step": 0.5},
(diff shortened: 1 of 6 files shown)


tlambert03:
going to merge this so that we can play with it in https://github.com/pymmcore-plus/pymmcore-widgets/pull/293

PR #355, opened 2024-06-05
feat: add the tensorstore option to the MDARunner 

fdrgsp:
@tlambert03  not sure what you think about it but in this PR I tried to find a way to add the `tensorstore` option to the `MDARunner` `_handler_for_path` method so we can pass something like `.my/path/ts.tensorstore.zarr` to the `run_mda` method and return a `TensorStoreHandler`.

Basically it assumes that the path for a `tensorstore` ends with either `.tensorstore` or with `.tensorstore.<driver>` (where `driver` is one of the supported `zarr`, `zarr3`, `n5` or `neuroglancer_precomputed` drivers). 

We search for the pattern `.tensorstore` in the provided path and, if it matches, we create a `TensorStoreHandler` with a `zarr` default driver if the path ends with `.tensorstore` or we get the driver form the suffix (for example, if the provided path is `.my/path/ts.tensorstore.n5`, we will use the driver `n5`).

### DIFF
--- a/src/pymmcore_plus/mda/_runner.py
+++ b/src/pymmcore_plus/mda/_runner.py
@@ -189,6 +189,7 @@ class MDARunner:
                 created automatically based on the extension of the path.
                 - `.zarr` files will be handled by `OMEZarrWriter`
                 - `.ome.tiff` files will be handled by `OMETiffWriter`
+                - `.tensorstore.zarr` files will be handled by `TensorStoreHandler`
                 - A directory with no extension will be handled by `ImageSequenceWriter`
             - A handler object that implements the `DataHandler` protocol, currently
                 meaning it has a `frameReady` method.  See `mda_listeners_connected`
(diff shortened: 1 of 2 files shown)


tlambert03:
i'm a bit hesitant to introduce what is effectively a more elaborate string-based API for selecting a Handler.  Passing a string should probably be used only when you care what the final file looks like (but *not* the exact implementation details of how pymmcore-plus actually writes the file).  That is, if you say `.ome.zarr` you get an ome-zarr structured file; you don't care whether tensorstore was used in the background or not (note: I recognize that that handler doesn't yet write ome format).

we already have a way for someone to request a more specific data handler, and that is to simply pass the handler, exactly as you want it.  So, this new API seems to exist mostly as a workaround for the MDAWidget to not have to construct a handler.  And I don't think it's worth adding a new string-based API just to avoid having the widget do that.  In fact, if the widget made its own handler, it would be even easier for it to pass that handler to things like the stack viewer (which would solve another one of your problems).  So, rather than this, why don't you just have the widget make the handler?  I know we avoided that at first, because there were only two options for handlers, but now might be a good time to revisit that decision

fdrgsp:
> So, rather than this, why don't you just have the widget make the handler?

closing because of https://github.com/pymmcore-plus/pymmcore-widgets/pull/298.

PR #351, opened 2024-05-14
feat: formalizing metadata during MDA

tlambert03:
this is a WIP PR implementing #350

Issue #350, opened 2024-05-14
discussion: metadata schema during MDA


tlambert03:
I'd like to begin nailing down a schema for metadata emitted during multi-dim acquisitions that consumers can rely on.  That is, I want to clarify the structure of the `meta` objects passed as the second argument to [`sequenceStarted`](https://pymmcore-plus.github.io/pymmcore-plus/api/events/#pymmcore_plus.mda.events._protocol.PMDASignaler.sequenceStarted) and the third argument to [`frameReady`](https://pymmcore-plus.github.io/pymmcore-plus/api/events/#pymmcore_plus.mda.events._protocol.PMDASignaler.frameReady).  This is tricky because some people will certainly want as much data as possible (which takes time) while others will want speed and minimal data on each frame.

These `meta` objects are currently dicts (but see below), and my plan is to:
1. have a `format` and `version` keys at the top level that define the structure of the metadata objects.  Versions will be X.Y, where `X` implies a breaking change (removal/renaming of a key or change in the type), while `Y` implies a backwards compatible change (addition of a non-required key).
2. allow users to set the `format` and `version` in the engine.  Something like
    ```python
    core.mda.engine.set_summary_metadata_format(format='full', version='1.0')`
    ```
    This way, if we later decide we need to make changes, any external consumers needn't be affected as long as they specify the format that they want.  (We'll just retain the ability to generate older versions indefinitely)

These formats will be accompanied by documentation that specifies the structure/schema that you can expect.

Some questions:

1. Does that general idea of a metadata format/version string sound good?
2. I'd *like* to depend on [msgspec](https://github.com/jcrist/msgspec), which is a really great high performance library for declaring structured schema, creating objects (faster than dicts), and encoding/decoding to/from json or msgpack (it will also make documenting these objects easier because it outputs to json schema, similar to pydantic).  It has no additional dependencies and has a great author.  Given that the final destination of most of this metadata will be serialization to disk, its [encoding/decoding performance](https://jcristharif.com/msgspec/benchmarks.html) will be nice.  So, first question: **does anyone mind if I take on that additional dependency**?
3. If we do use msgspec, **should the `meta` objects themselves remain `dict`**, with the "actual" metadata struct as the value of some `data` key:
    ```python
    meta = {
      "format": "some-string",
      "version": "1.0",
      "data": SomeStruct(),
    }
    ```
[...]

PR #358, opened 2024-06-22
feat: Formalize schema for metadata used by MDA (and elsewhere)

tlambert03:
this is a variant of #351.

I'm pulling back on the `msgspec.Struct` idea that I floated in #350, and am just using `TypedDict`.  For serialization, I  include a general `json_dumps()` function that can be used to serialize any of these dicts, regardless of whether they contain custom types like `MDAEvent` or `MDASequence`.

The current pattern for metadata implemented here is this:

- the second argument to `sequenceStarted` callbacks will always be a dict in the form of `pymmcore_plus.mda.metadata.SummaryMetaV1`:
    
    ```python
    class SummaryMetaV1(TypedDict, total=False):
        """Complete summary metadata for the system. Version 1.0."""
    
        format: Literal["summary-dict"]
        version: Literal["1.0"]
        datetime: str
        devices: Tuple[DeviceInfo, ...]
        system_info: SystemInfo
        image_infos: Tuple[ImageInfo, ...]
        config_groups: Tuple[ConfigGroup, ...]
        pixel_size_configs: Tuple[PixelSizeConfigPreset, ...]
        position: Position
        mda_sequence: NotRequired[useq.MDASequence]
        extra: NotRequired[Dict[str, Any]]
    ```

    - this dict can be created from a core instance using `pymmcore_plus.mda.metadata.summary_metadata`
    - here is a complete example of what that could look like after casing to builtins

    <details>

    <summary>Example <code>SummaryMetaV1</code></summary>

    ```python
        {
            "devices": [
                {
                    "label": "DHub",
                    "library": "DemoCamera",
                    "name": "DHub",
                    "type": "HubDevice",
                    "description": "DHub",
                    "parent_label": None,
                    "properties": [],
                    "child_names": [
                        "DCam",
                        "DWheel",
                        "DStateDevice",
                        ...,  # More children
                    ],
                },
                {
                    "label": "Camera",
                    "library": "DemoCamera",
                    "name": "DCam",
                    "type": "CameraDevice",
                    "description": "Demo camera",
                    "parent_label": "DHub",
                    "properties": [
                        {
                            "name": "AllowMultiROI",
                            "value": "0",
                            "data_type": "int",
                            "allowed_values": ["0", "1"],
                            "is_read_only": False,
                            "is_pre_init": False,
                            "sequenceable": False,
                        },
                        {
                            "name": "AsyncPropertyDelayMS",
                            "value": "2000",
                            "data_type": "int",
                            "allowed_values": [],
                            "is_read_only": False,
                            "is_pre_init": False,
                            "sequenceable": False,
                        },
                        ...,  # More properties
                    ],
                },
                {
                    "label": "Dichroic",
                    "library": "DemoCamera",
                    "name": "DWheel",
                    "type": "StateDevice",
                    "description": "Demo filter wheel",
                    "parent_label": "DHub",
                    "properties": [
                        {
                            "name": "ClosedPosition",
                            "value": "0",
                            "data_type": "int",
                            "allowed_values": [
                                "0",
                                "1",
                                "2",
                                "3",
                                "4",
                                "5",
                                "6",
                                "7",
                                "8",
                                "9",
                            ],
                            "is_read_only": False,
                            "is_pre_init": False,
                            "sequenceable": False,
                        },
                        {
                            "name": "Description",
                            "value": "Demo filter wheel driver",
                            "data_type": "str",
                            "allowed_values": [],
                            "is_read_only": True,
                            "is_pre_init": False,
                            "sequenceable": False,
                        },
                        ...,  # More properties
                    ],
                    "labels": [
                        "400DCLP",
                        "Q505LP",
                        "Q585LP",
                        "89402bs",
                        "State-4",
                        "State-5",
                        "State-6",
                        "State-7",
                        "State-8",
                        "State-9",
                    ],
                },
                ...,  # More devices
                {
                    "label": "Core",
                    "library": "",
                    "name": "Core",
                    "type": "CoreDevice",
                    "description": "Core device",
                    "parent_label": None,
                    "properties": [
                        {
                            "name": "AutoFocus",
                            "value": "Autofocus",
                            "allowed_values": ["", "Autofocus"],
                        },
                        {
                            "name": "AutoShutter",
                            "value": "1",
                            "allowed_values": ["0", "1"],
                        },
                        {
                            "name": "Camera",
                            "value": "Camera",
                            "allowed_values": ["", "Camera"],
                        },
                        {
                            "name": "ChannelGroup",
                            "value": "Channel",
                            "allowed_values": [
                                "",
                                "Camera",
                                "Channel",
                                "Channel-Multiband",
                                "LightPath",
                                "Objective",
                                "System",
                            ],
                        },
                        {
                            "name": "Focus",
                            "value": "Z",
                            "allowed_values": ["", "Z"],
                        },
                        {
                            "name": "Galvo",
                            "value": "",
                            "allowed_values": [""],
                        },
                        {
                            "name": "ImageProcessor",
                            "value": "",
                            "allowed_values": [""],
                        },
                        {
                            "name": "Initialize",
                            "value": "1",
                            "allowed_values": ["0", "1"],
                        },
                        {
                            "name": "SLM",
                            "value": "",
                            "allowed_values": [""],
                        },
                        {
                            "name": "Shutter",
                            "value": "White Light Shutter",
                            "allowed_values": ["", "LED Shutter", "White Light Shutter"],
                        },
                        {
                            "name": "TimeoutMs",
                            "value": "5000",
                            "allowed_values": [],
                        },
                        {
                            "name": "XYStage",
                            "value": "XY",
                            "allowed_values": ["", "XY"],
                        },
                    ],
                },
            ],
            "system_info": {
                "pymmcore_version": "0.10.3.dev21+gad932b1.d20240622",
                "pymmcore_plus_version": "0.10.3.dev21+gad932b1.d20240622",
                "mmcore_version": "MMCore version 11.1.1",
                "device_api_version": "Device API version 71, Module API version 10",
                "device_adapter_search_paths": [
                    "~/Library/Application Support/pymmcore-plus/mm/Micro-Manager-80d5ac1"
                ],
                "system_configuration": "~/Library/Application Support/pymmcore-plus/mm/Micro-Manager-80d5ac1/MMConfig_demo.cfg",
                "primary_log_file": "~/Library/Application Support/pymmcore-plus/logs/pymmcore-plus.log",
                "circular_buffer_memory_footprint": 250,
                "buffer_total_capacity": 0,
                "buffer_free_capacity": 0,
                "timeout_ms": 5000,
            },
            "image_info": {
                "bytes_per_pixel": 2,
                "current_pixel_size_config": "Res10x",
                "exposure": 10.0,
                "image_bit_depth": 16,
                "image_buffer_size": 524288,
                "image_height": 512,
                "image_width": 512,
                "magnification_factor": 1.0,
                "number_of_camera_channels": 1,
                "number_of_components": 1,
                "pixel_size_affine": [1.0, 0.0, 0.0, 0.0, 1.0, 0.0],
                "pixel_size_um": 1.0,
                "roi": [0, 0, 512, 512],
                "camera_device": "Camera",
                "multi_roi": [[], [], [], []],
            },
            "position": {"x": -0.0, "y": -0.0, "focus": 0.0},
            "config_groups": [
                {
                    "name": "Camera",
                    "presets": [
                        {
                            "name": "HighRes",
                            "settings": [
                                {"dev": "Camera", "prop": "Binning", "val": "1"},
                                {"dev": "Camera", "prop": "BitDepth", "val": "12"},
                            ],
                        },
                        {
                            "name": "LowRes",
                            "settings": [
                                {"dev": "Camera", "prop": "Binning", "val": "4"},
                                {"dev": "Camera", "prop": "BitDepth", "val": "8"},
                            ],
                        },
                        {
                            "name": "MedRes",
                            "settings": [
                                {"dev": "Camera", "prop": "Binning", "val": "2"},
                                {"dev": "Camera", "prop": "BitDepth", "val": "10"},
                            ],
                        },
                    ],
                },
                {
                    "name": "Channel",
                    "presets": [
                        {
                            "name": "Cy5",
                            "settings": [
                                {"dev": "Dichroic", "prop": "Label", "val": "400DCLP"},
                                {"dev": "Emission", "prop": "Label", "val": "Chroma-HQ700"},
                                {"dev": "Excitation", "prop": "Label", "val": "Chroma-HQ570"},
                                {
                                    "dev": "Core",
                                    "prop": "Shutter",
                                    "val": "White Light Shutter",
                                },
                            ],
                        },
                        {
                            "name": "DAPI",
                            "settings": [
                                {"dev": "Dichroic", "prop": "Label", "val": "400DCLP"},
                                {"dev": "Emission", "prop": "Label", "val": "Chroma-HQ620"},
                                {"dev": "Excitation", "prop": "Label", "val": "Chroma-D360"},
                                {
                                    "dev": "Core",
                                    "prop": "Shutter",
                                    "val": "White Light Shutter",
                                },
                            ],
                        },
                        ...,  # More channels
                    ],
                },
                ...,  # More config groups
            ],
            "pixel_size_configs": [
                {
                    "name": "Res10x",
                    "pixel_size_um": 1.0,
                    "pixel_size_affine": [1.0, 0.0, 0.0, 0.0, 1.0, 0.0],
                    "settings": [
                        {"dev": "Objective", "prop": "Label", "val": "Nikon 10X S Fluor"}
                    ],
                },
                ..., # More pixel size configs
            ],
            "format": "summary-dict-full",
            "date_time": "2024-06-22T15:46:57.561894",
            "version": "1.0",
            "mda_sequence": {
                "stage_positions": [
                    {"x": 0.0, "y": 0.0, "z": 0.0},
                    {"x": 1.0, "y": 1.0, "z": 1.0},
                ],
                "channels": [{"config": "DAPI"}, {"config": "FITC"}],
                "time_plan": {"interval": 0.01, "loops": 2},
                "z_plan": {"range": 0.3, "step": 0.1},
            },
        }
        ```

</details>

- the third argument to `frameReady` callbacks will always be a dict in the form of `pymmcore_plus.mda.metadata.FrameMetaV1`, which represents a very minimal amount of information:

    ```python
    class FrameMetaV1(TypedDict, total=False):
        """Metadata for a single frame. Version 1.0."""
    
        format: Literal["frame-dict"]
        version: Literal["1.0"]
        pixel_size_um: float
        camera_device: Optional[str]
        exposure_ms: float
        position: Position
        property_values: Tuple[PropertyValue, ...]
        runner_time_ms: float
        mda_event: NotRequired[useq.MDAEvent]
        hardware_triggered: NotRequired[bool]
        images_remaining_in_buffer: NotRequired[int]
        camera_metadata: NotRequired[Dict[str, Any]]
        extra: NotRequired[Dict[str, Any]]
    ```
    - this dict can be created from a core instance using `pymmcore_plus.mda.metadata.frame_metadata
    - here is an example of what that might look like after casing to builtins
    <details>
    
    <summary>Example <code>FrameMetaV1</code></summary>
    
    ```python
    {
        'exposure_ms': 10.0,
        'pixel_size_um': 1.0,
        'position': {'x': 222.0, 'y': 1.005, 'focus': 1.0},
        'camera_device': 'Camera',
        'property_values': [
            {'dev': 'Dichroic', 'prop': 'Label', 'val': '400DCLP'},
            {'dev': 'Emission', 'prop': 'Label', 'val': 'Chroma-HQ700'},
            {'dev': 'Excitation', 'prop': 'Label', 'val': 'Chroma-HQ570'},
            {'dev': 'Core', 'prop': 'Shutter', 'val': 'White Light Shutter'}
        ],
        'format': 'frame-dict-minimal',
        'version': '1.0',
        'mda_event': {
            'index': {'t': 0, 'p': 0, 'c': 0},
            'channel': {'config': 'Cy5', 'group': 'Channel'},
            'min_start_time': 0.0,
            'x_pos': 222.0,
            'y_pos': 1.0,
            'z_pos': 1.0,
            'sequence': {
                'axis_order': ['t', 'p', 'c', 'z'],
                'stage_positions': [{'x': 222.0, 'y': 1.0, 'z': 1.0}, {'x': 111.0, 'y': 0.0, 'z': 0.0}],
                'channels': [{'config': 'Cy5'}],
                'time_plan': {'interval': 1.5, 'loops': 2}
            }
        },
        'runner_time': 0.02322895801626146
    }
    ```
    
    </details> 

- while these dicts *may* have objects that are not natively serializable using `json.dumps` (such as `useq` objects or `datetime` objects, etc...), we provide a few functions that are guaranteed to serialize these objects, or cast them to (serializable form)
    - `pymmcore_plus.mda.metadata.serialize.json_dumps` - this function returns a JSON string in `bytes` 
    - `pymmcore_plus.mda.metadata.serialize.to_builtins` - this function returns a dictionary containing only objects that are serializable using `json.dumps`
- if `msgspec` is installed, it will be used, providing generally at least a 2-2.5x speedup over standard lib json
- I'll add clear docs explaining what the keys and value types are in a later PR (they are all defined in [`mda.metadata.schema`](https://github.com/pymmcore-plus/pymmcore-plus/pull/358/files#diff-60fd557aad8beab5d1b083f5ff34593bb2270533ad34ad0fc2188b0e72701acd)
- in a future PR, i may establish an API to allow users to further prune these objects (particularly the summary metadata).  But the objects defined here will always be a super-set of the final metadata.  If we do allow pruning, we'll use `typing.NotRequired` to mark fields that may be missing so that downstream consumers know that they should look-before-they-leap on those keys


tlambert03:
ok @wl-stepp, @marktsuchida, @dpshepherd,

this is where stuff landed with metadata (see #350 for initial discussion).  If you are so inclined, now would be a good time to peek at the first post above in this PR and let me know if you have any thoughts or concerns about what you see.

wl-stepp:
Looking good to me, very excited to get this. How would you imagine "custom" metadata to come in? Like EDA event scores for example?

tlambert03:
> How would you imagine "custom" metadata to come in? Like EDA event scores for example?

that's a good question:  should we put an additional `extra: dict[str, Any]` on both of these for user data?

wl-stepp:
> that's a good question: should we put an additional `extra: dict[str, Any]` on both of these for user data?

I feel like that would be a good idea. Will think about where should populate that.

wl-stepp:
Yes I would think in frameReady, but are there some writers that write the metadata directly with each frame or is it always aggregated? Would we be able to add metadata in another frameReady call in another thread for example and the writer could still save it? Maybe that's even something we should not consider and the metadata should stay about the caption of the frame only. Then we would have to think of another pattern for saving this information.
Same goes a little bit for SummaryMetadata for adaptive acquisitions. I see that is handled mostly in sequenceStarted, right? What if it would change during the acquisition? Would every writer take that into account?

tlambert03:
> are there some writers that write the metadata directly with each frame or is it always aggregated? ... Maybe that's even something we should not consider and the metadata should stay about the caption of the frame only. Then we would have to think of another pattern for saving this information.

yes, i don't see this PR and this specification as demanding anything about what a writer does with this info.  I think one of the issues that I see with metadata handling in (current) MMStudio and Java micromanager is that it attempts to standardize what *all* writers must do; and as a result of that, it's sort of stuck in a pattern where it *must* collect a large amount of metadata.  It makes sense of course: it's valuable to be able to say "anything written by micromanager will have X data"... but it does present legitimate constrictions on cases where you *don't* actually need to collect all of that information.

So, essentially, what we're establishing here is the "outer bounds" on what metadata can look like.  The full schema, if you will, of what metadata might look like at different stages of a sequence acquisition.  We leave open the possibility that different experimental needs will want to prune that down to more a stringent spec.  Writers, therefore, will have also have to deal with that possibility.  We can absolutely have a convention for the "standard" full-metadata, something like MMStudio currently has (that standard would likely just be the PR as it currently stands).  But to meet the broadest range of needs, I think we have to allow for the possibility that some writers will not be compatible with some metadata.  I think that's just fine (as long as you have a way up front to determine whether a given data handler is compatible with the metadata that a run is going to produce ... which we can have).

> Same goes a little bit for SummaryMetadata for adaptive acquisitions. I see that is handled mostly in sequenceStarted, right? What if it would change during the acquisition? Would every writer take that into account?

Summary metadata is sort of by definition a comprehensive state of things at the beginning of the experiment.  Frame metadata is for stuff that changes during the acquisition.  So, in an adaptive acquisition, anything that is changed during the course of the experiment would have to go in one or more of the frame metadatas.  Each frame metadata will have the MDAEvent that triggered it (that has its own user `metadata` field), and then the `FrameMetadata` `TypedDict`  spec here also has an `extra` field for you to put whatever you need in it.   As for writing that, the Tensorstore handler (which I would suggest is the best approach for adaptive acquisitions) will just write all the metadata it gets to disk.  So, as long as you get the metadata you need in any of `MDASequence.metadata`, `MDAEvent.metadata`, `SummaryMetadata['extra']` or `FrameMetadata['extra']`, then the writer will write it to disk :)

marktsuchida:
@tlambert03 As I said elsewhere, I like this approach of straightforwardly (though I imagine this took a lot of thinking) recording the available information without prematurely cooking it. Totally agree with what you said above about the problems with the MMStudio approach -- things that are not available or applicable should not show up at all.

I started jotting down some item-by-item comments, which I will get back to tomorrow. But here are some questions about the global structure:

- When there are multiple cameras acquiring simultaneously (multi-camera), would I be right to think that a `FrameMetaV1` applies to one image plane from one physical camera? (Bikeshed: would `PlaneMetaV1` or `ImagePlaneMetaV1` avoid the potential confusion with "frame" meaning a point in a time series (from cinematography -- as in "frame rate")?)

- I notice that `ImageInfo` is in `SummaryMetaV1` and not in `FrameMetaV1`. But some of the items in `ImageInfo` can be per-camera (such as `pixel_size_um` and `exposure[_ms]` which you already allow in `FrameMetaV1` but also `pixel_size_affine`, `roi` and a couple others) or may be allowed to be per-camera in the future (at least potentially, all of the fields). Could it make sense to have an `ImageInfo` also in `FrameMetaV1`, with the recommendation being that the unchanging fields be recorded in the summary's `image_info` and the plane-dependent values be recorded in the plane's `image_info`? (OTOH, it would be understandable if it is made a requirement that image size and pixel format be uniform. Perhaps there needs to be an `ImageInfo` limited to data shape, and a separate `PlaneInfo` or `PhysicalInfo`?)

- Some cameras generate arbitrary metadata fields for each frame of a sequence acquisition. These can be hardware-counted frame number, hardware-measured time, etc. It might be useful to separate these from the generic user-provided `extra` -- or maybe not. One argument for separating them (aside from clarifying provenance) is that they can be confusing to interpret if not known to originate from the hardware (because, e.g., the "image number" wouldn't necessarily have anything to do with the timepoint number of an MDA). Related to this, MMCore also adds `ElapsedTime-ms` (if not set by the camera; not to be confused with the cooked `ElapsedTime-ms` from acqEngine) and `ImageNumber` to sequence acquisition frames.

tlambert03:
thanks @marktsuchida, your eyes on this are *very* appreciated.

> would I be right to think that a FrameMetaV1 applies to one image plane from one physical camera

Yes, correct.  The creation and emission of frame metadata for a simple (non-sequenced) MDAevent happens [here](https://github.com/pymmcore-plus/pymmcore-plus/pull/358/files#diff-d2d22d3cf19ad0a98da516b776df8b54c2f902c9fb0ed070924eb1e80ecdba8aR265-R271), and indeed it yields a metadata for each channel_index/physical camera.  You raise a good point about the confusion between "frame" and "plane".  One minor issue with not using the word frame here is that the callback pattern `frameReady(image, event, meta)` also uses the word `frame`, though indeed it too would be emitted twice for a multi-channel camera.  Also potentially confusing.  Need to think about that

> I notice that ImageInfo is in SummaryMetaV1 and not in FrameMetaV1.  Could it make sense to have an ImageInfo also in FrameMetaV1, with the recommendation being that the unchanging fields be recorded in the summary's image_info and the plane-dependent values be recorded in the plane's image_info?

yeah, thanks for raising that.  I like the idea of putting a full `ImageInfo` in FrameMeta, and I also definitely like the concept (in general) that the frame meta is reserved for things that change. So that one can get the "full" ImageInfo by essentially doing `summary_meta['image_info'].update(frame_meta['image_info'])`.  Will play with that and may ping you directly for thoughts

> Some cameras generate arbitrary metadata fields for each frame of a sequence acquisition. It might be useful to separate these from the generic user-provided extra -- or maybe not

Yeah, currently those get "lost" after being retrieved [right here](https://github.com/pymmcore-plus/pymmcore-plus/pull/358/files#diff-d2d22d3cf19ad0a98da516b776df8b54c2f902c9fb0ed070924eb1e80ecdba8aR451).  Would be easy enough to just pass them through, and I do like the idea of creating a new unstructured key-value object in frame metadata for the camera

> Related to this, MMCore also adds ElapsedTime-ms (if not set by the camera; not to be confused with the cooked ElapsedTime-ms from acqEngine) and ImageNumber to sequence acquisition frames.

Yeah, I wanted to ask you about time stamps in general.  I see timing being measured/populated in many different places in micro-manager and it would be good to hear your thoughts on which should be used, and which may have been built on top of vestigial needs/concerns

marktsuchida:
> Yeah, I wanted to ask you about time stamps in general. I see timing being measured/populated in many different places in micro-manager and it would be good to hear your thoughts on which should be used, and which may have been built on top of vestigial needs/concerns

There is no clean story here, unfortunately. I think ElapsedTime-ms is the oldest one. It appears that the distinction between time measured in the camera hardware (which should have relative accuracy of a crystal oscillator) vs software (which has ms to 10ms accuracy depending on OS and various things, and is also delayed (and jittered) by buffering in the driver, even if assigned before going into the MMCore sequence buffer) was not thought to be important in the early days. So MMCore adds it using software time only if it is not set by the camera. It is meant to be the time since the "start" of the current sequence acquisition (the "start" is generally before the first frame, but the time lag is camera/driver-dependent). I suggest that this field be interpreted as software-generated, and it should be a future goal for MMDevice to define a separate field that guarantees hardware timestamping.

(acqEngine further adds ElapsedTime-ms to snapped images and adds the start offset for sequence-acquired images (IIRC). The result is useful for getting the approximate time, but I wouldn't trust it for anything requiring sub-few-seconds accuracy.)

Since Oct, 2022 (https://github.com/micro-manager/mmCoreAndDevices/commit/be99feb864c72650be20d24d98e0059bd2344ab3) the Core-generated ElapsedTime-ms uses C++'s `std::steady_clock`, so should not be affected by time zone changes and clock adjustments.

ElapsedTime-ms is probably used widely enough that we cannot easily deprecate it.

MMCore also adds a `g_Keyword_Metadata_TimeInCore` (= `"TimeReceivedByCore"`) tag (https://github.com/micro-manager/mmCoreAndDevices/commit/1e7cb324f1f6c4ce8c51917abb33c84278bd6f40). This is a formatted calendar datetime (ISO8601-ish), but its fatal flaw is that it is in local time (so information loss will occur during the autumn daylight time switch). I suspect it is not widely used (currently 1 use in our Java code), so we should probably replace this with a Unix timestamp in UTC, perhaps in nanoseconds (also reducing formatting overhead). Anybody reading this, don't start using it!

I do think it is reasonable for MMCore to add the (calendar and steady) times to frames, as long as it does not incur much overhead such as for formatting. People needing millisecond or better precision will always need to have the time recorded or generated in hardware (either the camera, or whatever device is triggering the camera).

There might be a use for an absolute steady timestamp (not relative to the start of a sequence acquisition). This can only be guaranteed to be consistent within a single process (though in practice it might be consistent until the computer is rebooted), but could still simplify a bunch of things.

Also, some cameras have the capability to record absolute timestamps (e.g., time since the camera was switched on).

The other related field added by MMCore is ImageNumber, which is just an index within a sequence acquisition, per camera label. I seem to remember that some cameras also generate something like this, but I don't remember if they used conflicting keys.

I guess I'll turn this into issues for MMCore :)

tlambert03:
ok yeah: i also saw the various places that `ElapsedTime-ms` is created and inserted.  Does seem like something with heavily overloaded semantics in a way.  For our purposes, there is a runner clock, that maintains a `t0` (the equivalent of AcqEngine's `correspondingEvent.acquisition_.getStartTime_ms()`, and stamps each event using [perf_counter](https://docs.python.org/3/library/time.html#time.perf_counter).  
So, it looks like the only additional thing we should have a proper place for is any intra-sequence timing as added by cameras.

> I do think it is reasonable for MMCore to add the (calendar and steady) times to frames

if so, where would one retrieve that info.  if it were inside of a sequenced image, then of course it can go in the circularbuffer metadata.  but otherwise?  would you have to add a `getImageTime` to match `getImage`?

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+]
+
+
+class PropertyInfo(TypedDict):
+    """Information about a single device property."""
+
+    name: str
+    value: Optional[str]
+    data_type: Literal["undefined", "float", "int", "str"]
+    allowed_values: Optional[Tuple[str, ...]]
```
marktsuchida, line 27: `NotRequired`?

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    is_read_only: bool
+    is_pre_init: bool
+    limits: NotRequired[Tuple[float, float]]
+    sequenceable: bool
```
marktsuchida, line 31: Since most properties are non-pre-init and non-sequenceable, it might be nice to omit these fields normally. Does that require `NotRequired[bool]`?

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+class DeviceInfo(TypedDict):
+    """Information about a specific device."""
+
+    label: str
+    library: str
+    name: str
+    type: str
+    description: str
+    properties: Tuple[PropertyInfo, ...]
+    parent_label: Optional[str]  # None on hub devices
```
marktsuchida, line 45: (nitpick) None on hub devices and devices that are not peripherals of a hub.

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    focus_direction: NotRequired[Literal["Unknown", "TowardSample", "AwayFromSample"]]
+    # hub device only
+    child_names: NotRequired[Tuple[str, ...]]
+    # stage or XYstage device only
+    is_stage_sequenceable: NotRequired[bool]
+    # stage device only
+    is_continuous_focus_drive: NotRequired[bool]
+    is_stage_linear_sequenceable: NotRequired[bool]
+    # camera device only
+    is_exposure_sequenceable: NotRequired[bool]
```
marktsuchida, line 59: SLMs might also have this capability, potentially if not currently.

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    library: str
+    name: str
+    type: str
+    description: str
+    properties: Tuple[PropertyInfo, ...]
+    parent_label: Optional[str]  # None on hub devices
+
+    # state device only
+    labels: NotRequired[Tuple[str, ...]]
+    # focus device only
```
marktsuchida, line 49: (for consistency) "focus device" and (non-XY) "stage device" are the same thing; see comments on items below.

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    focus_direction: NotRequired[Literal["Unknown", "TowardSample", "AwayFromSample"]]
+    # hub device only
+    child_names: NotRequired[Tuple[str, ...]]
+    # stage or XYstage device only
+    is_stage_sequenceable: NotRequired[bool]
+    # stage device only
+    is_continuous_focus_drive: NotRequired[bool]
+    is_stage_linear_sequenceable: NotRequired[bool]
```
marktsuchida, line 57: This linear sequencing thing is something that I once thought ASI stages could do and (sort of) implemented, but it ended up never working the way we expected. It's so obscure and probably unused that it might make sense to ignore it. Not that it harms anything to have the field.

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+
+class SystemInfo(TypedDict):
+    """General system information."""
+
+    pymmcore_version: str
+    pymmcore_plus_version: str
+    mmcore_version: str
+    device_api_version: str
+    device_adapter_search_paths: Tuple[str, ...]
+    system_configuration: Optional[str]
```
marktsuchida, line 70: Bikeshed: `hardware_configuration_file` (or `system_configuration_file`)?

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    primary_log_file: str
+    circular_buffer_memory_footprint: int
+    buffer_total_capacity: int
+    buffer_free_capacity: int
```
marktsuchida, line 74: I would use `sequence_buffer_size` (rather than "footprint"), possibly with a unit suffix (MB or bytes). "Circular buffer" is (actually _was_) an implementation detail.

`buffer_total_capacity` is that size divided by the image size so is redundant, and also I'm not sure the CMMCore method returns a meaningful value before the buffer is (re)allocated for the current image size. Consider leaving out.

`buffer_free_capacity` should be left out because it is a transient state that has more to do with what you were doing previously, and I can't think of a situation where it would be useful.

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    timeout_ms: int
+    # remaining_image_count: int
```
marktsuchida, line 76: I would leave out `remaining_image_count` at the summary level; it is mostly meaningless. For individual frames, when sequence-acquired, it is a useful diagnostic of buffer utilization.

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    continuous_focus_enabled: bool
+    continuous_focus_locked: bool
+    auto_shutter: bool
+
+
+class ImageInfo(TypedDict):
+    """Information about the current image structure."""
+
+    bytes_per_pixel: int
```
marktsuchida, line 85: I would consider adding a field that specifies the general category of pixel format, with options like "gray" and "ABGR". The main reason is to allow readers to reject any future pixel formats that it does not know how to interpret. Such future formats may include RGB (without the wasted channel), YUV (perhaps), or RAW (with a new field specifying the Bayer mask type).

Fully general pixel format specs (cf. GenICam Pixel Format Naming Convention) would be overkill, but by having even one category (could even be called "mm-classic", combining gray and ABGR), future extensions can be made without breaking readers.

Also, it is worth considering using bytes-per-component rather than bytes-per-pixel.

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    current_pixel_size_config: str
+    exposure: float
+    image_bit_depth: int
```
marktsuchida, line 88: Possibly `component_bit_depth` (there are RGB cameras that are 12 or 14-bit for each component).

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    image_buffer_size: int
+    image_height: int
+    image_width: int
+    magnification_factor: float
+    number_of_camera_channels: int
```
marktsuchida, line 93: Needs to be clarified in what case this would be ≠ 1.

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    number_of_components: int
```
marktsuchida, line 94: Maybe `components_per_pixel`?

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    pixel_size_affine: Tuple[float, float, float, float, float, float]
```
marktsuchida, line 95: This probably should not be a required field. Unfortunately we currently record the identity affine transform when really there is not a set transform, but it would be good to detect that and convert to "no affine transform".

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    pixel_size_um: float
+    roi: List[int]
+    camera_device: str
```
marktsuchida, line 98: Should this be `camera_label`?
But also it is possible to have a camera "label" without a physical camera of that name existing (not common, but if a camera emulates Multi Camera -- our LSM camera (OpenScan) does this).

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    multi_roi: NotRequired[Tuple[List[int], List[int], List[int], List[int]]]
+
+
+class Position(TypedDict):
+    """Represents a position in 3D space and focus."""
+
+    x: Optional[float]
+    y: Optional[float]
+    focus: Optional[float]
```
marktsuchida, line 107: The MMStudio MDA supports multiple XY stages and multiple Z stages. I suspect it will make sense to be able to record multiple stages' positions, even if there are primary XYZ coordinates that can be derived from them? The common case is motorized XYZ + piezo Z (+ piezo XY).

Also, calling the Z stage "focus" in position contexts is not always correct, because any single-axis stage (sometimes even rotary stages) could be used. Admittedly there may not be a correct answer here.

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+
+    name: str
+    settings: Tuple[PropertyValue, ...]
+
+
+class PixelSizeConfigPreset(ConfigPreset):
+    """A specialized group of device property settings for a pixel size preset."""
+
+    pixel_size_um: float
+    pixel_size_affine: Tuple[float, float, float, float, float, float]
```
marktsuchida, line 129: Again, affine transform should be optional.
(Unless you are willing to take (or already have taken) other measures to ensure that the correct scaling-only affine transform is always present based on the pixel size.)

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    """Complete summary metadata for the system. Version 1.0."""
+
+    devices: Tuple[DeviceInfo, ...]
+    system_info: SystemInfo
+    image_info: ImageInfo
+    config_groups: Tuple[ConfigGroup, ...]
+    pixel_size_configs: Tuple[PixelSizeConfigPreset, ...]
+    position: Position
+    mda_sequence: NotRequired[useq.MDASequence]
+    date_time: str
```
marktsuchida, line 149: Consider specifying that it must be UTC.

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+
+class ConfigGroup(TypedDict):
+    """A group of configuration presets."""
+
+    name: str
+    presets: Tuple[ConfigPreset, ...]
+
+
+class SummaryMetaV1(TypedDict, total=False):
+    """Complete summary metadata for the system. Version 1.0."""
```
marktsuchida, line 140: Looks like this contains enough information to recover the entire hardware config file, which is wonderful. If you haven't done so yet, it might be good to make sure this is the case.

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+
+    exposure_ms: float
+    pixel_size_um: float
+    position: Position
+    camera_device: Optional[str]
+    mda_event: NotRequired[useq.MDAEvent]
+    runner_time: NotRequired[float]
+    property_values: Tuple[PropertyValue, ...]
+    is_sequence_running: NotRequired[bool]
+    remaining_image_count: NotRequired[int]
```
marktsuchida, line 166: As noted above, this `remaining_image_count` could be a useful diagnostic.

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,169 @@
...
+    """Metadata for a single frame. Version 1.0."""
+
+    exposure_ms: float
+    pixel_size_um: float
+    position: Position
+    camera_device: Optional[str]
+    mda_event: NotRequired[useq.MDAEvent]
+    runner_time: NotRequired[float]
+    property_values: Tuple[PropertyValue, ...]
+    is_sequence_running: NotRequired[bool]
```
marktsuchida, line 165: Perhaps better to record whether the plane originated from a sequence acquisition (vs snap). If that is not feasible, not sure if this is needed.

marktsuchida (commented):
This is a somewhat raw braindump of my thoughts on individual fields. Happy to discuss more on each.

marktsuchida:
> > I do think it is reasonable for MMCore to add the (calendar and steady) times to frames
> 
> if so, where would one retrieve that info. if it were inside of a sequenced image, then of course it can go in the circularbuffer metadata. but otherwise? would you have to add a `getImageTime` to match `getImage`?

Well, that is a design bug in the snap API, that also needs to be addressed eventually. But the need is more critical for sequence acquisitions, because there is the possibility of images spending significant time in the sequence buffer, often more than the upstream delay up to the driver. By the above sentence I just meant that it makes sense to keep (or replace/improve) the current timestamps added by MMCore.

src/pymmcore_plus/mda/metadata/schema.py
tlambert03: yes!  #359

src/pymmcore_plus/mda/metadata/schema.py
tlambert03: `sequence_buffer_size_mb` 👍

tlambert03: done

src/pymmcore_plus/mda/metadata/schema.py
tlambert03: added a comment, and renaming to `number_of_camera_adapter_channels` ...though it might even be nice to remove this altogether as it's such an implementation detail of MM

src/pymmcore_plus/mda/metadata/schema.py
tlambert03: a more generic model could be:


```python
class StagePosition(TypedDict):
    device_label: str
    position: float | tuple[float, ...] 

positions: list[StagePosition]
```

or (some variant like that)

src/pymmcore_plus/mda/metadata/schema.py
tlambert03: what about a timezone aware timestamp, that doesn't necessarily have to be UTC offset 0:

```python
In [10]: datetime.datetime.now(datetime.UTC).astimezone().isoformat()
Out[10]: '2024-06-26T19:35:09.963496-04:00'

In [11]: datetime.datetime.now(datetime.UTC).isoformat()
Out[11]: '2024-06-26T23:35:18.202662+00:00'
```

marktsuchida: That should be fine since it has a bijection to/from UTC. Nice idea for human-friendliness!

src/pymmcore_plus/mda/metadata/schema.py
```diff
@@ -0,0 +1,206 @@
...
+    name: str
+    presets: Tuple[ConfigPreset, ...]
+
+
+class SummaryMetaV1(TypedDict, total=False):
+    """Complete summary metadata for the system. Version 1.0."""
+
+    devices: Tuple[DeviceInfo, ...]
+    system_info: SystemInfo
+    image_infos: Tuple[ImageInfo, ...]
```
marktsuchida, line 181: Can you remind me how the correct index for this is determined for a given plane/frame? Or is that still TBD?

marktsuchida (commented):
My thoughts on the `ImageInfo` fields.

src/pymmcore_plus/mda/metadata/schema.py
tlambert03: TBD ... or, you would use the `camera_device` label in FrameMetadata to look up stuff in SummaryMetadata `(next(x for x in summary['image_infos'] if x.get('camera_label') == 'thing'))`

tlambert03:
ok, thanks everyone for all the input on this.  I think this is in a good place (and I think i've saturated my ability to evaluate this PR), so I'm going to merge after tests pass, and then follow up with any downstream tweaks/consequences in later PRs.
Will post the link to the docs page here once it's live

PR #365, opened 2024-07-03
feat: guard against premature call to getInstalledDevices

tlambert03:
similar to https://github.com/micro-manager/mmCoreAndDevices/pull/476 but on the python side

it's important not to call `getInstalledDevices` before calling `initializeDevice`, otherwise you may never be able to query the peripherals of a given device (even after later initialization)


tlambert03:
this won't be necessary with https://github.com/micro-manager/mmCoreAndDevices/pull/476 merged.

PR #366, opened 2024-07-05
refactor: use only pydantic2 syntax

tlambert03:
in preparation for the next useq-schema release, this uses only pydantic v2 syntax


tlambert03:
not sure why pre-commit.ci isn't triggering, but this is linting fine locally.  merging

PR #386, opened 2024-10-25
Add a hint about the mmcore build-dev command

ctrueden:
When `mmcore build` fails due to unsupported platform (e.g. Linux), give a hint that the `mmcore build-dev` command also exists to build Micro-Manager from source.


ctrueden:
Sorry, I don't know if this change is a "feat" or a "docs" or what. :shrug:

tlambert03:
thanks for taking the time @ctrueden, you're an open source hero :)

PR #392, opened 2024-11-08
perf: don't query position on triggered acquisition frames by default, add `include_frame_position_metadata` flag

### LINKED ISSUE #379: Acquisition stops temporarily while it catches up with frames
Related to this discussion: https://imagesc.zulipchat.com/#narrow/stream/442785-pymmcore.5B-plus.5D/topic/questions and to the micromanager-gui project as well.

Specifically: "Is it possible to run an acquisition in micromanager-gui without displaying the images? We've noticed that when performing fast acquisitions, often the acquisitions stop at the right time (say, a "fast" 20 or 50ms exposure acquisition for 5s - we see the led come on and off for those 5s) but then it waits until all the images have been displayed before starting on the second channel. We're not too interested in seeing all the images live (maybe the first for each channel would be good for sanity checking the focus is holding, but not too important) so if live display could be disabled, that would be great for us!"

The issue presents itself even if the acquisition is started as a script. I'm attaching here the full console log from the script run!
[console_log.txt](https://github.com/user-attachments/files/17166476/console_log.txt)

tlambert03:
This PR makes it so that the MDAEngine doesn't query stage positions during sequenced/triggered acquisitions by default.  This can be a rather slow call, depending on the stage, and ultimately was at the heart of @simonecoppola's performance issues seen in https://github.com/pymmcore-plus/pymmcore-plus/issues/379 (thank you **SO** much Simone for debugging that).

this PR adds:

```python
MDAEngine:
    include_frame_position_metadata: Literal[True, False, "unsequenced-only"] = "unsequenced-only"
```

which gate whether or not we ask for position.  This is not a "generalizable" solution, as was discussed at the end of #358 (in that it doesn't add a more generic pattern for pruning metadata as desired), but it gets something down for now.

note that `FrameMetaV1.position` now becomes an `NotRequired[Position]`

Note: an elaboration here, should anyone *really* want position information during sequenced acquisitions, is to check whether the `SequencedEvent` itself is changing XYZ during the sequence (right around [here](https://github.com/pymmcore-plus/pymmcore-plus/blob/0f39fb39ab039330986f4b49b7231bd64142ed8a/src/pymmcore_plus/mda/_engine.py#L426-L438))... and if not, just query the position once at the beginning, and use the same position metadata for all the frames (rather than re-querying non-changing information)

fixes #379

### DIFF
--- a/src/pymmcore_plus/mda/_engine.py
+++ b/src/pymmcore_plus/mda/_engine.py
@@ -3,11 +3,7 @@ from __future__ import annotations
 import time
 from contextlib import suppress
 from itertools import product
-from typing import (
-    TYPE_CHECKING,
-    NamedTuple,
-    cast,
-)
+from typing import TYPE_CHECKING, Literal, NamedTuple, cast
 
 from useq import HardwareAutofocus, MDAEvent, MDASequence
 
@@ -27,12 +23,15 @@ from ._protocol import PMDAEngine
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator, Sequence
+    from typing import TypeAlias
 
     from numpy.typing import NDArray
 
     from pymmcore_plus.core import CMMCorePlus
 
     from ._protocol import PImagePayload
+
+    IncludePositionArg: TypeAlias = Literal[True, False, "unsequenced-only"]
 
 
 class MDAEngine(PMDAEngine):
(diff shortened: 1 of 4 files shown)


tlambert03:
merging.  @simonecoppola tested and said it would fix https://github.com/pymmcore-plus/pymmcore-plus/issues/379

Issue #376, opened 2024-09-23
State of writers for MultiCamera acquisitions


wl-stepp:
Dear all,

I wanted to see what the current state for acquiring and storing MultiCamera acquisitions. I see that it runs Ok if I set it up correctly from the Engine and Runner side. I don't get two channels in the saved Zarrs however.
I think there are two problems here:

- event.index is the same for both

{'format': 'frame-dict', 'version': '1.0', 'runner_time_ms': 291.50379999919096, 'camera_device': **'HamamatsuHam_DCAM'**, 'property_values': ({'dev': 'CSUW1-Filter Wheel 1', 'prop': 'Label', 'val': 'Cy3'}, {'dev': 'CSUW1-Filter Wheel 2', 'prop': 'Label', 'val': 'DAPI'}, {'dev': 'DA TTL State Device', 'prop': 'State', 'val': '5'}, {'dev': 'Core', 'prop': 'Shutter', 'val': 'State Device Shutter'}, {'dev': 'CondenserTurret', 'prop': 'Label', 'val': '7-Closed'}), 'exposure_ms': 42.954, 'pixel_size_um': 0.065, 'position': {'x': 2344.8, 'y': -3700.2000000000003, 'z': 1968.02}, 'mda_event': MDAEvent(index=mappingproxy(**{'c': 0}**), channel=Channel(config='Dual-DAPI-Cy3', group='3-Channel'))}

 {'format': 'frame-dict', 'version': '1.0', 'runner_time_ms': 291.50379999919096, 'camera_device': **'HamamatsuHam_DCAM-1'**, 'property_values': ({'dev': 'CSUW1-Filter Wheel 1', 'prop': 'Label', 'val': 'Cy3'}, {'dev': 'CSUW1-Filter Wheel 2', 'prop': 'Label', 'val': 'DAPI'}, {'dev': 'DA TTL State Device', 'prop': 'State', 'val': '5'}, {'dev': 'Core', 'prop': 'Shutter', 'val': 'State Device Shutter'}, {'dev': 'CondenserTurret', 'prop': 'Label', 'val': '7-Closed'}), 'exposure_ms': 42.954, 'pixel_size_um': 0.065, 'position': {'x': 2344.8, 'y': -3700.2000000000003, 'z': 1968.02}, 'mda_event': MDAEvent(index=mappingproxy(**{'c': 0}**), channel=Channel(config='Dual-DAPI-Cy3', group='3-Channel'))}

[...]

PR #394, opened 2024-11-10
Dual cam writer

wl-stepp:


wl-stepp:
I have thought about #376 a little more. If I understood the discussion there correctly, the logic should be limited to the writer. The problem there is, that I have not found a way to determine the cameras that will actually be imaging from SummaryMetaV1 in sequenceStarted and there is no access to the core, which limits what preparation is possible and what to do in new_store. I also think it might be beneficial to have that information in the metadata. I have added a special key for that, but it could also go into image_infros for example.

Just a starting point for deeper discussions maybe. I will put something along these lines into place for one of our applications.

tlambert03:
Thanks for starting this @wl-stepp, it's very helpful to see an implementation.  I have a number of thoughts, but no convictions yet on the right way to do this.  The whole thing is obviously quite hacky, due to the MultiCamera device itself being hacky.  I'm ok with hacky approaches to deal with hacky implementations, but I'd like to "silo" them as much as possible.  some unordered thoughts:

> I have not found a way to determine the cameras that will actually be imaging from SummaryMetaV1 in sequenceStarted and there is no access to the core

- one is actually able to derive most of the core state from the SummaryMetadata using the 'Core' device.  So here's how you could get `active_cameras` from summary meta.  (this assumes the multicam device, so would need an additional conditional to bail on the normal case)

    ```python
    def get_real_cam_devices(meta: SummaryMetaV1) -> tuple[str, ...]:
        devices = meta["devices"]
        # find the core device
        core = next(d for d in devices if d["label"] == "Core")
        # same as getCameraDevice
        cam = next(p for p in core["properties"] if p["name"] == "Camera")
        cur_cam_device = next(d for d in devices if d["label"] == cam["value"])
        # same hacky solution we do in `core.getPhysicalCameraDevice`
        return tuple(
            p["value"]
            for p in cur_cam_device["properties"]
            if p["name"].startswith("Physical Camera") and p["value"]
        )
    ```
- one thing that would make this all easier, and which we should add to the SummaryMeta, is to know how many channels each camera offers.  basically, we should uncomment [this line](https://github.com/pymmcore-plus/pymmcore-plus/blob/0f39fb39ab039330986f4b49b7231bd64142ed8a/src/pymmcore_plus/metadata/functions.py#L177-L178) so that each camera_label in `image_infos` shows how many channels it expects
- Stepping back, it seems to me that we don't *really* need to know any of the names of camera devices though, so this may all be a distraction.  What we really need to do is have TensorStoreHandler (or any other handler) be ready to handle the case where it receives a `frameReady` call that comes along with an `event. index` that has already been seen before.  For example if it gets a second `{'t': 0, 'c': 0}` event, then it should be smart enough to expand the "virtual" channel axis accordingly.  This could be done one of two ways:
    - the `new_store` could greedily, in *every* case, just create an additional singleton axis ready to handle the edge case of two images for the same channel.  So, for example, shape would go from `(nT, nC)` to `(nT, nC, 1)`.  And then, if we ever receive more than 1 image for the same MDAEvent.index, we would grow that singleton dimension with `self._store.resize`.  (the main point here is that the Tensorstore store itself doesn't need to have the same literal shape as the final dataset.  It's easy to `.squeeze()` and `.reshape()` things as needed to expose what feels like a standard shape to a consumer while still having the degrees of freedom needed internally to handle multiple circumstances.
    - **probably even better: we just *always* internally use this mode where all the Tensorstore handler does is dump each frame to a new chunk on disk** (essentially a growing 3D array where every dimension besides XY is intertwined in the third dim).  And then provide nD views onto that linear array of chunks.  That might not immediately make sense, so I'll try to expand on that idea soon.
- as an aside: i don't think the writer should modify the MDAEvent like is done here.  I think the MDAEvent should be treated like the *request* that was given to the microscope, and if we modify it before writing it in the metadata to disk, then it's hard to know what the original request was.  Admittedly, it's confusing here, because the "shape" of the original requested MDASequence is not longer matching the shape of the final data.  But i think that mismatch should be be discernible somehow (rather than obscured) in the final data written to disk.  This might actually be a good issue to raise in useq-schema: perhaps we need a way to indicate that the `event.channel` will actually result in multiple *actual* channels.

tlambert03:
> probably even better: we just always internally use this mode where all the Tensorstore handler does is dump each frame to a new chunk on disk

expanding on this idea a bit more.  what I'm proposing is that we just always live in [`_nd_storage = False` mode](https://github.com/pymmcore-plus/pymmcore-plus/blob/0f39fb39ab039330986f4b49b7231bd64142ed8a/src/pymmcore_plus/mda/handlers/_tensorstore_handler.py#L124-L129), where the layout of chunks on disk are in a single (flat) directory.  Zarr makes it pretty easy to present a bunch of on-disk chunks in any sort of multidimensional view, just be changing the mapping of key (representing some high dimensional index) to value (actual bytes on disk).  And then, at the end of the experiment, it would also be easy enough to re-arrange the chunks into a multi-directory representation once all the experimental dimensions are revealed:

while writing the experiment
```
├── 0
├── 1
├── ...
└── N
```

reshaped on-disk representation after experiment
```
└── 0
    └── 0
        ├── 0
        │   ├── 0
        │   └── 1
        └── 1
            └── 0
```

wl-stepp:
Thanks for those thoughts. Helps a lot in seeing things clearer. I get the separation between MDAEvent as 'command' for the microscope and a separate description of the resulting data. 

> probably even better: we just always internally use this mode where all the Tensorstore handler does is dump each frame to a new chunk on disk 
Yes, that would help a lot with the adaptive acquisitions where we can't know at all what events will come in. Isolation also might be possible into just frameReady...
Any thoughts on how we would handle that on the viewer side? Would a viewer working during acquisition need an indexed view into the structure in order to represent the different axes well?

tlambert03:
> that would help a lot with the adaptive acquisitions

yeah, that was the original motivation for that storage scheme (https://github.com/pymmcore-plus/pymmcore-plus/pull/348)

> Any thoughts on how we would handle that on the viewer side? 

i'm thinking of a combination of things... and this generally applies to all data backends for ndv

1. the data store (in this case tensorstorehandler) offers two methods, very much like the xarray API: 
    - `dims`: returns an ordered sequence of the number of dimensions
    - `coords`: returns a mapping of dimension name to coordinates (this can be case to `.shape`)
1. the data store also has a way to signal that the general shape (be it dims or coords) have changed.  In this case TensorStoreHandler has a way to say "a new dimension has been added" or "the shape of an existing dimension has changed"
1. regardless of how the data chunks are actually stored on disk (in this case, a flat/linear sequence of XY image chunks), the data store is responsible for offering the viewer an nD "view" of the data (so that the viewer can use the same query method for any type of data store to say "give me these indices along these dimensions")
1. that view is responsible for converting the high-dimensional data request/query into actual locations on disk

wl-stepp:
This is what I can come up with
```
            index = dict(event.index)
            while frozenset(index.items()) in self._frame_indices.keys():
                if index['c'] in self._index_jump_map.keys():
                    index['c'] = self._index_jump_map[index['c']]
                    continue
                self._index_jump_map[index['c']] = self._axis_max.get('c', 0) + 1
                index['c'] = self._index_jump_map[index['c']]
            self._frame_indices[frozenset(index.items())] = ts_index
```
But it depends on always all 'low' channel indexes coming in and them coming in the same order, so not ideal. The other thing I can think of would be to add an additional (temporary) 'folded_c' index into _frame_indices that gets integrated into the 'c' axis in finalize metadata for example.

wl-stepp:
So far pretty simple to pull out into a subclass:
https://github.com/LEB-EPFL/pymmcore-eda/pull/4

tlambert03:
sorry for the long delay on this.   Mostly dragging my feet because A) I think you've got a decent workaround working locally?  and B) there are a few things about the data-storage pattern that i'm unhappy with at the moment, so every time i think about this one i'm reminded of the other issues that I feel like need to be solved (mostly, synchronicity of the callbacks, and the general difference between deterministic and non-deterministic sequences).

wl-stepp:
Yes, locally we have something working decently. I think those are all points that come into play here and a more general approach to the matter would be great, feel free to close if to specific

PR #399, opened 2024-11-15
initial proof of concept cached events

simonecoppola:


tlambert03:
hey @simonecoppola, with the refactoring of sequencing done in #400 and other stuff, do you know if we could still benefit from caching here?  Are you still doing something like this on your end?  Happy to have something like this in here if there is benefit to be had

simonecoppola:
I haven't had a chance to test things further but from what I remember from the initial tests I did was that this was better performance wise, though I'm not sure why that would be the case as we're doing the same thing? 
It works quite well for us when running MDAs with basically zero gap between the various acquisitions, though I have just today discovered a pretty bad bug where after the MDA, if one starts a Live view, the system just keeps repeating the sequence😅 so it clearly needs some thinking/work...

I will do a few more tests to compare them and update you in the next few days!

PR #400, opened 2024-11-17
perf: better sequenced event building

### LINKED ISSUE #396: Sequencing/triggering across different channels could be improved.
The logic in `can_sequence_events` currently looks like this:

https://github.com/pymmcore-plus/pymmcore-plus/blob/0f39fb39ab039330986f4b49b7231bd64142ed8a/src/pymmcore_plus/core/_sequencing.py#L238-L252

that could be probably be improved:

- we likely don't need to prevent sequencing across config groups as is currently done
- we *also* probably don't need to return `_nope` when `core.isPropertySequenceable` is False, provided that the value of the property isn't changing between the two events
- so this should be refactored to convert the two channel configs into a set of properties *whose values change* from one event to the next, and then only check that subset of properties for sequencability

thanks @marktsuchida

tlambert03:
This is a big PR that hopefully improves a number of issues with building sequenced/triggered acquisitions.

- closes #396 by ensuring that we only try to sequence values and properties that are actually changing (allowing non-sequencable properties to be a part of the configuration as long as they don't change)
- hopefully speeds up the process of building sequences by only checking things like `is*Sequenceable` and `get*SequenceMaxLength` once for the entire iterator.
- make it easier for someone to see what the runner would do when sequencing events with `list(pymmcore_plus.core.iter_sequenced_events(core, events))`
- makes a `mocks.py` module with a `MockSequenceableCore` that can be used to test this stuff without needing the underlying devices to actually support sequencing.
- ~fixes a bug (for which there's no issue) when using triggered acquisition and passing a Queue based (event driven) iterator to `MDARunner.run()`.  In this PR, if an iterator is directly passed, then we don't attempt to merge any events (since it could result in a deadlock while we look ahead in the events).~  (that's fixed now in #419)
- make `SequencedEvent` public in `pymmcore_plus.core`

for anyone having a look at this PR, the easiest entry point is probably `pymmcore_plus.core.iter_sequenced_events`.  Which takes an `Iterable[MDAEvent]`, and returns another `Iterable[MDAEvent]`, where some of those events may actually be SequencedEvents

```python
from pymmcore_plus.core import iter_sequenced_events
from useq import MDASequence

core = CMMCorePlus.instance()
events = MDASequence(time_plan={'interval': 0, 'loops': 10}, z_plan={'range': 4, 'step': 0.5})
list(iter_sequenced_events(core, events))
```


simonecoppola:
hi @tlambert03 , i've just run a few mda acquisitions with this and everything seems to be working ok! is there any more specific testing you'd like me to perform?

tlambert03:
how easy is it to create a little code that demonstrates the problem?

simonecoppola:
Here it is!

```python

from useq import MDASequence
import time

mmc = CMMCorePlus.instance()
cfg_path = r"path\to\config.cfg"
mmc.loadSystemConfiguration(cfg_path)

mmc.setConfig("Channel","Ex470Em519-26")

print("Running first live acquisition")
# First test that things are ok
mmc.startContinuousSequenceAcquisition(100)
# pause execution for 30s

time.sleep(15)
mmc.stopSequenceAcquisition()

# Now generate and then run mda acquisition
sequence = MDASequence(
    channels=[{"config": "Ex470Em519-26", "exposure": 50}, {"config": "Ex550Em595-33", "exposure": 50}],
    time_plan={"interval": 0, "loops": 50},
    axis_order="ct",
    keep_shutter_open_across=('t',)
)


print("Running sequenced MDA")
mmc.mda.engine.use_hardware_sequencing = True
acquisition = mmc.run_mda(sequence)
acquisition.join() # wait until it's done to continue

print("Running second live acquisition")

# Now test again continuous sequence acquisition
mmc.setConfig("Channel", "Ex470Em519-26")

# First test that things are ok
mmc.startContinuousSequenceAcquisition(100)

# pause execution for 30s
time.sleep(15)
mmc.stopSequenceAcquisition()

```


What I'm observing is that during the first 15s of `ContinuousSequenceAcquisition` the channel is just the 470 one. Then the MDA happens as it should. When I then start the second `ContinuousSequenceAcquisition`, instead of the channel staying as the 470 one, it alternates between that and the other one - repeating the MDA sequence.

Hope this makes sense - can provide a video of the microscope as well in case!

simonecoppola:
Sorry for the double comment. Just to follow up on that, I though I'd share the config file as well:
[openframe_config.txt](https://github.com/user-attachments/files/18732631/openframe_config.txt)

tlambert03:
thanks!

PR #406, opened 2024-11-24
feat: support slm_image from useq v0.6

hinderling:
Made some progress! Tested with MightexPolygon1000 DMD. Submitting this pull request so we have a place for discussion, also there are some issues with `useq-schema` for which I'd be grateful for help.

With the current code, one can set `slm_image.data` to:
- True / False single bool: All pixels are set to ON/OFF, looking up the default value which is hard coded in `pymmcore-plus` for some devices.
- 0-255 single int: sets all pixels to this value (these are both running `setSLMPixelsTo`)
- [w,h,(rgb)] np.array of `u8ints`: sets DMD pixels to image (images use `setSLMImage`)
- [w,h,(rgb)] np.array of `bool`: sets DMD pixels to image, replacing True values with the hardcoded default value.

Example to set all pixels to ON state:
```python
event_slm_on = MDAEvent(slm_image=SLMImage(data=True))
mmc.mda.engine.setup_event(event_slm_on)
```

Issues:
- 1) There is an error in `useq._base_model`, `_non_default_repr_args` when comparing `np.arrays`. I just skipped this test for now, there is also a comment (`consider using model_copy directly`) - how should we proceed with this?
- 2) when using a mda_engine, slm_image is only applied if there is a single event in the list, see example below:

Works:
```python
events = []
for slm_image_data in [0]:
    slm_image = SLMImage(data = slm_image_data,device=slm_dev)
    events.append(useq.MDAEvent(exposure=500,slm_image=slm_image))

thread = mmc.run_mda(events)
thread.join()

events = []
for slm_image_data in [255]:
    slm_image = SLMImage(data = slm_image_data,device=slm_dev)
    events.append(useq.MDAEvent(exposure=500,slm_image=slm_image))

thread = mmc.run_mda(events)
thread.join()
```

Doesn't work:
```python
events = []
for slm_image_data in [0,255]:
    slm_image = SLMImage(data = slm_image_data,device=slm_dev)
    events.append(useq.MDAEvent(exposure=500,slm_image=slm_image))

thread = mmc.run_mda(events)
thread.join()
```

Other TODO:
- [x] Decide if/where to hardcode the SLM ON-pixel values.
- [x] Test with RGB
- [ ] I don't have a DMD with a builtin light source, so it's hard to test the exposure time and DMDs that require `displaySLMImage`. The polygon1000 sets pixels immediately when calling `setSLMImage` or `setSLMPixelsTo`. Will check if this is implemented in the genericSLMDevice.
- [ ] test with other DMDs
- [ ] fix device name lookup


tlambert03:
woot!  excited to take a look

tlambert03:
> There is an error in useq._base_model, _non_default_repr_args when comparing np.arrays. I just skipped this test for now, there is also a comment (consider using model_copy directly) - how should we proceed with this?

ah yes... the ever-annoying cannot test equality of numpy arrays :joy: ... let's fix that in useq schema.  there is some prior art for this, but for now just open an issue there

tlambert03:
btw, quick tip:  rather than

```python
thread = mmc.run_mda(events)
thread.join()
```

just do `mmc.mda.run(events)`   (the unthreaded version)

src/pymmcore_plus/mda/_engine.py
```diff
@@ -237,6 +249,57 @@ def setup_single_event(self, event: MDAEvent) -> None:
         if event.z_pos is not None:
             self._set_event_z(event)
 
+        if event.slm_image is not None:
```
tlambert03, line 252: some suggestions to simplify this a bit:

```python
        if event.slm_image is not None:
            try:
                # Get the SLM device
                if not (
                    slm_device := event.slm_image.device or self._mmc.getSLMDevice()
                ):
                    raise ValueError("No SLM device found or specified.")

                # cast to numpy array
                slm_array = np.asarray(event.slm_image)
                # if it's a single value, we can just set all pixels to that value
                if slm_array.ndim == 0:
                    value = slm_array.item()
                    if isinstance(value, bool):
                        on_value = _SLM_DEVICES_PIXEL_ON_VALUES.get(slm_device, 1)
                        value = on_value if value else 0
                    self._mmc.setSLMPixelsTo(slm_device, int(value))
                elif slm_array.size == 3:
                    # if it's a 3-valued array, we assume it's RGB
                    r, g, b = slm_array.astype(int)
                    self._mmc.setSLMPixelsTo(slm_device, r, g, b)
                elif slm_array.ndim in (2, 3):
                    # if it's a 2D/3D array, we assume it's an image
                    # where 3D is RGB with shape (h, w, 3)
                    if slm_array.ndim == 3:
                        if not slm_array.shape[2] == 3:
                            raise ValueError(
                                "SLM image must be 2D or 3D with 3 channels (RGB)."
                            )
                    # convert boolean on/off values to pixel values
                    if slm_array.dtype == bool:
                        on_value = _SLM_DEVICES_PIXEL_ON_VALUES.get(slm_device, 1)
                        slm_array = np.where(slm_array, on_value, 0).astype(np.uint8)
                    self._mmc.setSLMImage(slm_device, slm_array)

            except Exception as e:
                logger.warning("Failed to set SLM Image: %s", e)
```

- because `SLMImage`implements `__array__`, we can just cast the entire object to a numpy array, and I think that's a good pattern (it leaves open the possibility that the object itself can be involved in how it is cast to an array in the future.
- here i cast the whole thing to an array before checking all of the various data-type options.

tlambert03 (commented):
good start @hinderling!  thank you

I wonder whether the reason you might not be getting hits with multiple events has to do with sequencing?  It would depend on your config... but you can run  `list(mmc.mda.engine.event_iterator(events))` to see whether the engine is combining events into a `SequencedEvent` for you... and if so, then you'd need to implement `setup_sequenced_event` as well.

question for both you and @marktsuchida: is there any good way to test SLMdevices?  Or should we be creating a mock core that just behaves as if it was loaded with an SLM device?

src/pymmcore_plus/mda/_engine.py
tlambert03: and then probably move that whole try block into a new method `_set_event_slm` like the other event fields

hinderling: Thanks for your input!!!

> the reason you might not be getting hits with multiple events has to do with sequencing?

That was my guess as well! Will investigate tomorrow.

> is there any good way to test SLMdevices?

This is quite a bottleneck in developing for me, as the current genericSLM only works on Windows machines.
My idea was to implement it as a [pydevice](https://github.com/micro-manager/pymmdevice/issues/2#issue-2380395678), but never got around to finish this. Here's the [current state](https://github.com/micro-manager/mmCoreAndDevices/compare/main...hinderling:mmCoreAndDevices:slm_device).

I started messing around with the useq np.array issue [here](https://github.com/hinderling/useq-schema/tree/slm_image), let me know if this could work and if I should open a pull request, or just write an issue. I'm very unfamiliar with pydantic.

tlambert03: great thanks.  

> This is quite a bottleneck in developing for me, as the current genericSLM only works on Windows machines. My idea was to implement it as a pydevice.

I'll be honest, I'm slightly apprehensive about pydevices on the python side.  It makes a ton of sense for those wanting to stick with the Java ecosystem and write device adapters in python that the (C++) core can manage.
But it would be a somewhat unnatural way to write a device adapter in python, if we're *already* in python; and it creates yet another interface that one needs to learn and support.

By all means, you should absolutely keep working on it for the benefit of the C/Java stuff if you prefer to write device adapters in python instead of C.

But for the sake of testing on the python side, it would be much easier for us to create a `unittest.mock.Mock` or some other python object (without going back through the C side) that does everything we need without a real device attached.

> let me know if this could work and if I should open a pull request, or just write an issue.

using `np.array_equal` somewhere is indeed the correct approach, but using ` arbitrary_types_allowed=True` is not (there are better ways).  Please open an issue with a small bit of code that demonstrates the problem

hinderling: Perfect, thanks!! Will do.


If there is a way in pymmcore-(plus?) to cleanly add a mock SLM (and other) devices without having to deal with the C++ layer at all, that would be fantastic! Would much prefer that route. 


In the next step, my usecase would be to write a very basic camera+dmd simulator that allows us to actually influence the image acquisition (e.g. load an image from disk and multiply it with the SLM image), so that we can test feedback experiments completely in silico.

At the moment I have implemented this in a custom acquisition engine, but in the best case we would just change the cfg file to switch between emulation and real microscope.

hinderling:
- Tested with [#5563fc0 ](https://github.com/pymmcore-plus/useq-schema/pull/207), works!

- For now I just disabled sequencing if two slm_images are non-equal (again issues with np.arrays). What's your opinion on proper SLM sequencing? I'm afraid this could lead to a [rabbit hole](https://sourceforge.net/p/micro-manager/mailman/micro-manager-general/thread/1581116812587-0.post%40n2.nabble.com/), some DMDs working with `DisplaySLMImage` while some require `CommandTrigger`. Can we leave this for another time? Same with exposure, I don't have a good way of testing this.

Sample code for affine transformation calibration, works great! Much cleaner than what we had before. 
```python
p0, p1, p2 = ([456,380],[304,760],[608,760])

exposure = 500 
events = []
for p in [p0,p1,p2]:
    img_p = np.zeros((slm_height,slm_width)).astype(np.uint8)
    img_p[p[1],p[0]] = 255
    event_p = MDAEvent(slm_image=SLMImage(data=img_p,device=slm_dev),exposure=exposure)
    events.append(event_p)  

#disconnect previous callbacks
mmc.mda.events.frameReady.disconnect()

@mmc.mda.events.frameReady.connect
def new_frame(img: np.ndarray, event: MDAEvent):
    #find max value x/y
    max_x = np.argmax(img.max(axis=0))
    max_y = np.argmax(img.max(axis=1))
    plt.imshow(img,cmap='gray_r')
    plt.scatter(max_x,max_y,color='#00000000',marker='o',edgecolor='red')
    plt.show()

mmc.mda.run(events)
```

src/pymmcore_plus/mda/_engine.py
```diff
@@ -567,6 +580,42 @@ def _set_event_z(self, event: MDAEvent) -> None:
...
+                if slm_array.ndim == 3:
+                    if not slm_array.shape[2] == 3:
+                        raise ValueError(
+                            "SLM image must be 2D or 3D with 3 channels (RGB)."
+                        )
+                # convert boolean on/off values to pixel values
+                if slm_array.dtype == bool:
+                    on_value = _SLM_DEVICES_PIXEL_ON_VALUES.get(slm_device, 1)
+                    slm_array = np.where(slm_array, on_value, 0).astype(np.uint8)
+                self._mmc.setSLMImage(slm_device, slm_array)
```
tlambert03, line 614: I think we *may* want to call `self._mmc. displaySLMImage(slm_device)` here:

cc @marktsuchida, I'm not entirely sure when/whether one is supposed to call `displaySLMImage`.  Should we be doing it here immediately after setting it?  I looked through both the clojure and AcqEng and didn't see any calls to it.  The only thing I found was [here in SLM.java](https://github.com/micro-manager/micro-manager/blob/2244df2cc6e29e12221ff8c86702b9d32dc808d7/libraries/Projector/src/main/java/org/micromanager/projector/internal/devices/SLM.java#L167-L177), where pretty much every call to `setSLMImage` is followed immediately by `displaySLMImage`

src/pymmcore_plus/mda/_engine.py
```diff
@@ -567,6 +580,42 @@ def _set_event_z(self, event: MDAEvent) -> None:
         correction = self._z_correction.setdefault(p_idx, 0.0)
         self._mmc.setZPosition(cast("float", event.z_pos) + correction)
 
+    def _set_event_slm_image(self, event: MDAEvent) -> None:
+        try:
```
tlambert03, line 584: for reasons of typing, we either need to "recheck" the event here:

```python
if not event.slm_image:
    return
```
or cast it with `typing.cast` if we're sure this method will never be called with a null slm_image.

```python
slm_image = cast('useq.SLMImage', event.slm_image)
```

the first option is safer, and probably fine.  let's go with that

src/pymmcore_plus/mda/_engine.py
```diff
@@ -567,6 +580,42 @@ def _set_event_z(self, event: MDAEvent) -> None:
...
+                    if not slm_array.shape[2] == 3:
+                        raise ValueError(
+                            "SLM image must be 2D or 3D with 3 channels (RGB)."
+                        )
+                # convert boolean on/off values to pixel values
+                if slm_array.dtype == bool:
+                    on_value = _SLM_DEVICES_PIXEL_ON_VALUES.get(slm_device, 1)
+                    slm_array = np.where(slm_array, on_value, 0).astype(np.uint8)
+                self._mmc.setSLMImage(slm_device, slm_array)
+
```
tlambert03, line 615: somewhere we also need to call `if slm_image.exposure is not None: self._mmc.setSLMExposure(slm_device, slm_image.exposure)`

src/pymmcore_plus/mda/_engine.py
tlambert03: when I run this all locally, using a computer monitor as an SLM, I definitely don't see anything when running the example code that @hinderling provided in the post above unless i manually call `displaySLMImage` here.  seems fine to do it here, but we could also delay that to `exec_event` as well if there's any reason to wait

src/pymmcore_plus/mda/_engine.py
tlambert03: see #407 as a related possibility

src/pymmcore_plus/core/_sequencing.py
```diff
@@ -277,6 +278,21 @@ def _nope(reason: str) -> tuple[bool, str] | bool:
     elif cur_length >= core.getExposureSequenceMaxLength(cam_dev):  # pragma: no cover
         return _nope(f"Camera {cam_dev!r} {max_len=} < {cur_length=}")
 
+    # SLM
+    try:
+        if e1.slm_image != e2.slm_image:
```
tlambert03, line 283: ah, we could also make this easier in useq_schema as well.  one should be able to do an equality check without fear :). will make issue

src/pymmcore_plus/core/_sequencing.py
```diff
@@ -277,6 +278,21 @@ def _nope(reason: str) -> tuple[bool, str] | bool:
...
+            return _nope("SLMImage sequences are not supported yet.")
+    except ValueError:
+        if not np.array_equal(e1.slm_image, e2.slm_image):
+            return _nope("SLMImage sequences are not supported yet.")
+        # slm_dev = core.getSLMDevice()
```
tlambert03, line 288: I'm find with the above for now.  you can uncomment this and we'll just leave it as unimplemented

tlambert03 (commented):
good stuff!  I'm happy with how this is going.  I tried it locally and it worked nicely (once I added a call to `displaySLMImage` as mentioned).  I'll think about how we can test this, and may push directly to your branch

src/pymmcore_plus/mda/_engine.py
hinderling: I've added it to `exec_single_event` and `exec_sequenced_event`. If there is an exposure time set, the countdown would otherwise already start in the `setup_event`, my hope is to minimize camera/dmd exposure offset. Could you check if on your setup the image gets displayed now, even for sequenced acquisitions? 
Here I'm not sure what to do: Do we need to re-trigger `DisplaySLMImage` for every frame in a sequenced acquisition? 

Called when executing a single/sequenced event:
```python
def _exec_event_slm_image(self, event: MDAEvent) -> None:
    if not event.slm_image:
        return
    if slm_device := event.slm_image.device or self._mmc.getSLMDevice():
        try:
            self._mmc.displaySLMImage(slm_device)
        except Exception as e:
            logger.warning("Failed to set SLM Image: %s", e)
```

If two slm_images are the same, the events are sequenceable. Current check:
```python
if e1.slm_image != e2.slm_image:
    return _nope("SLMImage sequences are not supported yet.")
```

if `DisplaySLMImage` needs to be re-triggered for every acquisition, we could change it to:
```python
    if e1.slm_image is None or e2.slm_image is None:
        return _nope("SLMImage sequences are not supported yet.")
```
This would break however any sequencing with an SLM image, even for DMDs that don't need re-triggering. 
With the current implementation for me displaying `slm_img_0, slm_img_0, slm_img_0, slm_img_0` is faster than `slm_img_0 ,slm_img_1 ,slm_img_0, slm_img_1`, as the first sequence can be sequenced.

I think my confusion is stemming from not understanding sequencing. If we acquire a burst of 10 images, in a sequence, how many times is `exec_sequenced_event` executed?

src/pymmcore_plus/mda/_engine.py
hinderling: added first option

src/pymmcore_plus/mda/_engine.py
hinderling: Added at end of `_set_event_slm_image`.
Default for my DMD is 0 (display forever), which I would NOT overwrite with the camera exposure if no SLM exposure is set.

src/pymmcore_plus/core/_sequencing.py
hinderling: fearless equality check implemented and tested! works with updated version of useq.

src/pymmcore_plus/core/_sequencing.py
hinderling: removed comments

src/pymmcore_plus/mda/_engine.py
tlambert03: yeah, i agree.  let's leave out that fallback until we understand it better.  Basically `slm_image.exposure` is there if someone wants to call `setSLMExposure`, but we don't perform any other magic 👍

tlambert03:
couple more things need to be fixed here:

- change the dependencies in `pyproject.toml` to `useq-schema >=0.6.2`  to indicate that we have a new minimum requirement
- please also change it in `.pre-commit-config.yaml`, so that the typing checks are updated as well.

src/pymmcore_plus/mda/_engine.py
```diff
@@ -567,6 +587,58 @@ def _set_event_z(self, event: MDAEvent) -> None:
...
+        except Exception as e:
+            logger.warning("Failed to set SLM Image: %s", e)
+        if event.slm_image.exposure:
+            try:
+                self._mmc.setSLMExposure(event.slm_image.exposure)
```
tlambert03, line 629: `setSLMExposure` also needs the `slm_device` argument.  So you either need to move this up into the try/except block, or move the call to `if not (slm_device := event.slm_image.device` out above the try block; otherwise `slm_device` will potentially be unbound here (if we hit an exception above)

src/pymmcore_plus/mda/_engine.py
hinderling: Seems to work correctly with genericSLM device now! Still not clear what happens to SLMs where SLMExposure actually has an effect. Images are displayed forever on Polygon1000 and genericSLM, no matter the SLMExposure.

tlambert03:
just took another look and this is just about ready to go.  I added some tests in [6061bbd](https://github.com/pymmcore-plus/pymmcore-plus/pull/406/commits/6061bbdb2c9197f10e9ff6ce2e85afe8800fcbaa).  I did it by mocking the important core methods that actually need a device, just to assert that they've actually been called with the values we expect; that's all we need to test for anyway, since it's not our job to test the actual device adapters themselves

tlambert03:
hey @hinderling, I had a chance to look at the names in some real device adapters and updated this slightly in https://github.com/pymmcore-plus/pymmcore-plus/pull/406/commits/50e368c849b9786f1475a06255de80e8820ce6a9 will merge once tests pass.  Thanks again for your contribution here!

tlambert03 (approved):

PR #407, opened 2024-11-25
feat: Unicore - a unified core object that manages both C++ devices and python side devices in the same runtime

tlambert03:
@marktsuchida, here's me playing around with the ability to write python devices, and use them side by side with the existing core.  See `example.py` for how the final end-user api might look (the idea is that, you basically just use the core API exactly as you would have, but the core first checks to see whether a device label belongs to a python device, before d deferring to the CMMCore), and `_pycore.py` for the implementation (it's only XYStageDevices for now, and I don't think it would be necessary to have complete device coverage for this to be partially useful)

I'm *suuuuper* averse in general to establishing new interfaces, which is of course what is being done here... so the primary question for me here is exactly what the python side object should look like, and what the general "messaging" is around how and when/whether it should be used.  At the same time, the general question of "how can I just slightly extend the C-side devices with a little custom code on the python side" comes up all the time, and it's frustrating not to have an answer for that. (this week, it came up with both @hinderling and @jacopoabramo)

Obviously, this touches the same general space of the recently added pydevice in mmCoreAndDevices, which I think was primarily designed to allow people to write devices in python, and use them in the Java-based micromanager ecosystem.  Of course: python code is python code, and I absolutely think we should consider whether the same modules could be used in both cases (it would be great if the same code could be used).  There's a lot of subtleties to discuss there though.

------------------

*autogenerated summary* 

This pull request introduces several significant changes to the `pymmcore_plus` library, focusing on enhancing device management and error handling, adding new features, and improving code organization. The most important changes include the ability to use combined C and Python devices.

### New Features and Examples:
* Added a new example `examples/unicore.py` demonstrating how to create and use a custom Python stage device with UniCore. This example shows how to integrate Python devices with existing C++ devices in the `pymmcore_plus` library.

### Device Management:
* Introduced `PyDeviceManager` in `src/pymmcore_plus/experimental/unicore/_device_manager.py` to manage the loading, initialization, and unloading of Python devices. This class provides methods for device lifecycle management and ensures devices are correctly initialized and shutdown.
* Added the `create_proxy` function and related classes in `src/pymmcore_plus/core/_proxy.py` to create restricted proxy objects for `CMMCorePlus` instances. This allows for controlled access to core attributes and events, enhancing security and modularity.

### Error Handling:
* Improved error handling in the `loadDevice` method of `CMMCorePlus` by adding the `_load_error_with_info` method. This method provides more detailed error messages when device loading fails, helping to diagnose issues more effectively. [[1]](diffhunk://#diff-ef27d81e1fb28d6095de5e5db5411c37b8d780decea194ec6a393834793b0cddL353-R366) [[2]](diffhunk://#diff-ef27d81e1fb28d6095de5e5db5411c37b8d780decea194ec6a393834793b0cddL377-R383)

### Code Enhancements:
* Updated `src/pymmcore_plus/core/_constants.py` to include a new `Boolean` property type, adding support for boolean properties in the `PropertyType` enum.
* Added the `create` class method to `PropertyType` in `src/pymmcore_plus/core/_constants.py`, allowing for the creation of `PropertyType` instances from various input types, improving flexibility and usability.

These changes collectively make it easier to integrate and manage Python devices alongside C++ devices.


jacopoabramo:
Wasn't there a PR approved some time ago where python device adapters could be added in the core and proprties were defined via python properties?

jacopoabramo:
> I'm suuuuper averse in general to establishing new interfaces, which is of course what is being done here...

I mean, from the example you've shown there's nothing that already exists in pymmcore. Or am I missing something? To me, "new interface" would mean a new set of functions to control the movement of a stage or a new set of functions to control camera acquisition in a way that currently is not available in the core - which I believe this draft doesn't cover

hinderling:
@jacopoabramo 

> Wasn't there a PR approved some time ago where python device adapters could be added in the core and proprties were defined via python properties?

[PyDevice](https://github.com/micro-manager/mmCoreAndDevices/tree/main/DeviceAdapters/PyDevice)? Advantage is that it works with Java uManager, disadvantage is that it creates its own Python runtime, so when using it with pymmcore you get `Pymmcore Python interpreter` -> `C++` -> `PyDevice Python interpreter`. Also it was quite difficult to set up in my limited experience (e.g. pointing it to use the right Python VEnv).

tlambert03:
@marktsuchida, I made some updates and I think this is ready for something of a review.  I haven't yet worked on the base Device object (so properties and initialization and stuff isn't yet considered), but the higher level patterns are taking shape and it would be good to get your eyes on it.  See the `test_unicore` file for basic usage ... and the `unicore/_unicore.py` file for the logic of how we dispatch to python or C++ devices.

@jacopoabramo, if you have time to take a look, I'd also value your feedback

jacopoabramo:
Ok I managed to give a relative look to the PR (maybe not as in depth as I should be doing but I think I got the gist of the attempt).

I like it quite a lot, and there should be a discussion to setup about this approach.

In my very very very humble opinion, integrating motorized devices is probably the easiest thing to do; if I can even be so bold, it could be an attempt to present an unified interface for everything that is motorized to a certain degree.

What I imagine is that stages, motorized filter wheels and microscopy objectives could go in the same category as devices that perform step-specific movements, and the API could be adapted in a way that the input for a `move` call could accept either a set of coordinates (absolute or relative) or alternatively a specific string value equivalent to one supported by the other categories.

A similar approach could be set for galvo and SLMs which (although I'm unfamiliar with the latter) would sum up in performing some sort of scan.

But I understand that this is too disruptive and it'll just cause headaches to keep compatibility with existing hardware, so it's probably best if this is taken care of by some higher-level logic.

I'm curious on your idea of how to approach the camera. If memory serves right, the core has a shared memory buffer and with the multi camera adapter you could put images for different streams and pop them. Would this be kept also for possible new device adapters?

tlambert03:
Thanks for your feedback @jacopoabramo.  Very much appreciate you taking a look.

> But I understand that this is too disruptive and it'll just cause headaches to keep compatibility with existing hardware, so it's probably best if this is taken care of by some higher-level logic.

Yeah.  I do agree with many of your points here!  But a lot of this wades into what I meant by being averse to establishing new interfaces.  For now, the goal here is simply to expose an interface that directly mirrors the C-side interface so that people can *extend* the C++ devices with python devices.  But at this point I'm not trying to step back and reconsider device interface design.  I do definitely agree though that some higher level logic could perhaps provide an even simpler interface that marries the commonalities across device types!  That would be fun experiment, perhaps in another repo?

> I'm curious on your idea of how to approach the camera. If memory serves right, the core has a shared memory buffer and with the multi camera adapter you could put images for different streams and pop them. Would this be kept also for possible new device adapters?

I haven't really spent much time thinking about how to manage camera memory buffers yet.  I would say that the *goal* would be to have some solution on the python side that solves (or at least provides a workaround for eager programmers) many of the existing headaches surrounding multi-camera and/or performance bottlenecks.  That *includes* solutions that would allow the camera drivers to somehow directly write to disk, and i'll probably use python-acquire as a test case.

tlambert03:
This PR is pretty much done and I'm quite happy with the pattern.  Though verbose, I think it's pretty readable, and rather feature complete for the base device concept, properties, sequencing, and XY stages.

 It implements:

- a Unicore object that manages both python and C++ devices
- full support for properties, including sequencing
- support for property caching and device thread safety, similar to MMCore
- a sequenceable device base class.
- XYStage device as an initial implementation

I know it's a huge amount to have to review, so I'll leave it up here for a while, in case anyone wants to weigh in.  I'll branch off of this PR to implement support for the other device base classes.

jacopoabramo:
I would be up for experimenting with it for my new project. Is there some documentation already available on how this approach works?

Also happy holidays!

tlambert03:
The example is the best documentation at the moment, and the abstract base classes for each device are relatively self documenting (you won't be able to instantiate it without implementing all the needed methods)

There are only stage devices in this pr.  Others to follow

tlambert03:
https://github.com/tlambert03/pymmcore-plus/blob/pydevice-core/examples/unicore.py

tlambert03:
i'm going to merge this since it's been here for a while now.  It will remain in the `experimental` namespace.  Note that most of the devices (beyond stages) still need to have adapters added.  Those will come in later PRs, but if you stumble upon this PR and are waiting for a specific `DeviceType` to be added to the unicore, please comment here or open an issue :)

hinderling:
Heyo! Super cool to see this merged!!
Is there any documentation on Unicore? The blob linked above is not accessible anymore:
> https://github.com/tlambert03/pymmcore-plus/blob/pydevice-core/examples/unicore.py

I'm working on a [simulation for closed-loop optogenetic control](https://github.com/hinderling/opto-loop-sim), to test our smart microscopy pipelines completely in silico, this would require a fake python camera/slm device. Also this repo (https://github.com/pymmcore-plus/mda-simulator) seems close to what I wan't to be doing -- but would be fantastic ofc if one can switch from simulated/actual hardware just by changing the .cfg file. Would Unicore be the way to go here?

PR #413, opened 2024-12-02
feat: proof-of-principle using pymmcore-nano

tlambert03:
been playing around with nanobind bindings for pymmcore: https://github.com/pymmcore-plus/pymmcore-nano

this demonstrates that it's very nearly at feature parity with `pymmcore`... couple threading issues remaining

Note: i'm not entirely sure if/when we'll actually use it.  but it's there to be played with


tlambert03:
woohoo!  This is now all working.  there are a couple of tests in pymmcore-widgets that are failing.  most because return types of `tuple` have been converted to return types of `list` by using nanobind.  but the gist of the tests are still working.  there might be one thing with shutter state not working though ... need to look

PR #414, opened 2024-12-07
Unicore SLMDevice

tlambert03:
just a stub for now


hinderling:
Screenshot 2025-06-05 at 11 45 45 AM

Got an SLM implementation working! Cells move towards the light :) Should I try making a pull request, or will you be faster implementing it yourself anyways?
One issue I ran into was accessing the current SLM mask from the simulation, as there is no `getSLMImage`. Now using `displaySLMImage` to return the current image (instead of `none`) -- is there a proper way to do this?

tlambert03:
Actually @hinderling, maybe don't waste your time on the PR,  this was more complete than I thought.  If you want to have a quick look to make sure I didn't miss anything important, that would be great, then we can merge

tlambert03:
What device type next? :)

hinderling:
state & shutter device to simulate fluorescence imaging? also curious if there are advantages of using an image processor device vs. our current implementation with frameReady callbacks for real-time segmentation?

tlambert03:
> state & shutter device to simulate fluorescence imaging?

👍 

> advantages of using an image processor device vs. our current implementation with frameReady callbacks for real-time segmentation?

no not really.  A couple general notes:

1. a C++ imageProcessorDevice will *not* work with a python-side Camera Device.  (TLDR; both the SWIG pymmcore wrapper and the current pymmcore-nano implementation are not polymorphic, so the C++ CMMCore methods will never call back into python side Unicore code.  This means that we can't just implement a python camera device and have the C++ core use it as it would a regular C++ camera device, grabbing frame buffers from it and passing those buffers into C++ image processors)
2. I don't intend to re-implement ImageProcessorDevice any time soon on the Unicore side.  After talking with @marktsuchida, my understanding is that that API is a bit limited as is, and was originally added to be able to do simple shape-maintaining operations like reflections.  It's not that you *couldn't* do more with it, but the breadth of what could be done with python-side image processing isn't well covered by the MMCore image processor API.  And so yeah: I'm inclined to stick with purely python side image processing logic.
3. That's not to say we don't want to consider some sort of *structured* python-side logic.  Simply using `frameReady` works for an end user, but it's hard to coordinate multiple composable operations (for example, to share some aspect of processing before writing and/or viewing).  So that's definitely something we could consider adding.  But for now, just "take the image and do whatever you want in python" is the current state

hinderling:
> where the C-layer knows about the python devices

Super cool!!!

> the slm methods now optionally take no label

We have small convenience workaround currently: when loading the hardware config, we create a python microscope object that contains hardware specific helper functions / deals with microscope specific quirks. We use this also to store/access the "current" slm label (in quotations because we never change the slm after init, we also don't have any scopes with multiple slms).

PR #428, opened 2025-02-05
fix!: fix focus direction enums

### LINKED ISSUE #427: Bug: inconsistency in FocusDirection values leads to error when loading config file
@taebong tried to open a file written by MMStudio that had focusDirection = -1 and got an error in pymmcore.  Based on [the enum in MMDeviceConstants.h](https://github.com/micro-manager/mmCoreAndDevices/blob/056f4481c13c9ca06047c358277ba81032eb0faf/MMDevice/MMDeviceConstants.h#L269-L273) I would have expected it to be 0, 1, 2:

```python
In [2]: import pymmcore

In [3]: pymmcore.FocusDirectionAwayFromSample
Out[3]: 2

In [4]: pymmcore.FocusDirectionTowardSample
Out[4]: 1

In [5]: pymmcore.FocusDirectionUnknown
Out[5]: 0
```

but it looks like CMMCore [shifts all those enums](https://github.com/micro-manager/mmCoreAndDevices/blob/056f4481c13c9ca06047c358277ba81032eb0faf/MMCore/MMCore.cpp#L2063-L2074) and [so too does MMStudio](https://github.com/micro-manager/micro-manager/blob/17fd59e100c5ae104972d1a69b4fb9705090a939/mmstudio/src/main/java/org/micromanager/internal/hcwizard/Device.java#L415-L423)

we need to decide how to unify.  probably should just adopt the shifted convention and change the FocusDirection  enum value

tlambert03:
fixes #427 

pymmcore's `FocusDirectionUnknown`, `FocusDirectionTowardSample`, and `FocusDirectionAwayFromSample` constants expose the `MM::FocusDirection...` values (0,1,2) defined in [`MMDeviceConstants.h`](https://github.com/micro-manager/mmCoreAndDevices/blob/056f4481c13c9ca06047c358277ba81032eb0faf/MMDevice/MMDeviceConstants.h#L269-L273).  However, MMCore does not directly use those enum values, but rather [maps them to (-1, 0, 1)](https://github.com/micro-manager/mmCoreAndDevices/blob/056f4481c13c9ca06047c358277ba81032eb0faf/MMCore/MMCore.cpp#L2063-L2074)

so we shouldn't pass on the `pymmcore` constants, but rather declare our own

### DIFF
--- a/src/pymmcore_plus/core/_constants.py
+++ b/src/pymmcore_plus/core/_constants.py
@@ -209,10 +209,15 @@ class PortType(IntEnum):
     HIDPort = pymmcore.HIDPort
 
 
+# NB:
+# do *not* use `pymmcore.FocusDirection...` enums here.
+# the MMCore API does not use the device enums (which is what pymmcore exposes)
+# but instead translates MM::FocusDirectionTowardSample into a different number:
+# https://github.com/micro-manager/mmCoreAndDevices/tree/MMCore/MMCore.cpp#L2063-L2074
 class FocusDirection(IntEnum):
-    Unknown = pymmcore.FocusDirectionUnknown
-    TowardSample = pymmcore.FocusDirectionTowardSample
-    AwayFromSample = pymmcore.FocusDirectionAwayFromSample
+    Unknown = 0
+    TowardSample = 1
+    AwayFromSample = -1
     # aliases
     FocusDirectionUnknown = Unknown
     FocusDirectionTowardSample = TowardSample


tlambert03:
(technically a breaking change... but i don't think anyone will notice)

tlambert03:
cc @taebong.  please run `pip install -U pymmcore-plus` to get the newest version, which should fix your FocusDirection issue when loading configs from MMStudio

taebong:
Thanks!

PR #437, opened 2025-02-16
feat: extend object-oriented device API

tlambert03:
this adds `pymmcore_plus.core.Device` subclasses for each `DeviceType`, with device-type specific methods:

for example:

```python
In [1]: core = CMMCorePlus()

In [2]: core.loadSystemConfiguration()

In [4]: xy = core.getDeviceObject('XY')

In [5]: xy.position
Out[5]: (-0.0, -0.0)

In [6]: xy.position = (10, 20)

In [7]: xy.position
Out[7]: (10.004999999999999, 19.995)

In [10]: xy.getXYPosition()  # standard API alias
Out[10]: [10.004999999999999, 19.995]

In [12]: xy.isSequenceable()
Out[12]: False
```

cc @marktsuchida 

`core.getDeviceObject` also becomes type-safe.  If you know the type of the device for the label you're passing, you can pass either the `pymmcore_plus.core.Device` subclass, or the `DeviceType` enum and your IDE will know what subtype you're getting (and an exception will be raised if you were wrong about the device type)

Screenshot 2025-02-16 at 1 43 56 PM

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ ci:
 
 repos:
   - repo: https://github.com/crate-ci/typos
-    rev: v1
+    rev: v1.31.1
     hooks:
       - id: typos
         args: [--force-exclude]
--- a/src/pymmcore_plus/core/__init__.py
+++ b/src/pymmcore_plus/core/__init__.py
@@ -1,10 +1,14 @@
 __all__ = [
     "ActionType",
+    "AutoFocusDevice",
     "CFGCommand",
     "CFGGroup",
     "CMMCorePlus",
+    "CameraDevice",
     "ConfigGroup",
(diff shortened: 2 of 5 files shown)


tlambert03:
tried to add as much test coverage as possible... but some lines are hard to test with the available test config

marktsuchida:
LGTM except that this doesn't have the pump devices yet (could be added later, of course).

tlambert03:
> LGTM except that this doesn't have the pump devices yet (could be added later, of course).

true,  a release of https://github.com/micro-manager/pymmcore/pull/131 would be handy for that

marktsuchida:
> true, a release of [micro-manager/pymmcore#131](https://github.com/micro-manager/pymmcore/pull/131) would be handy for that

Done!

PR #439, opened 2025-02-20
feat: add pixel calibration routine

tlambert03:
Has no public API at the moment... but this is generically useful even in the absence of a GUI.

The measures pixel size and rotation (full affine)


tlambert03:
thanks for the heads up!  I'll give you a quick example of how to test this once I've finalized it

PR #441, opened 2025-02-21
feat: mock sample with space awareness (WIP)

tlambert03:
for working on #439, I really need some simulated data that responds to the state of the core (moving with the stage, aware of camera parameters, etc...). This is a proof of principle PR implementing such a framework.

https://github.com/user-attachments/assets/01633487-2be1-48e8-9cf8-7ee443e76eb0

### DIFF
--- a/src/pymmcore_plus/core/events/__init__.py
+++ b/src/pymmcore_plus/core/events/__init__.py
@@ -6,7 +6,7 @@ from ._protocol import PCoreSignaler
 from ._psygnal import CMMCoreSignaler
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


tlambert03:
I've updated this and moved it to experimental.    It's entirely opt in, works pretty well, and is tested. so I might merge this.   

I added support for both pillow and opencv (which is faster).  

full example in `examples/simulate_sample`

tlambert03:
@marktsuchida ... maybe a quick look at the general patterns?  with an eye towards forward compatibility with better things you're making?  (it's possible that there just isn't much API overlap between a c++ device and a core patcher, but just in case)

PR #451, opened 2025-03-25
test: fixes for pymmcore v11.5

tlambert03:
the new version of pymmcore changes a  few things that break tests.  This fixes it.  see also https://github.com/micro-manager/pymmcore/pull/131


tlambert03:
the only failing tests now needs to be fixed on the pymmcore-widgets side: https://github.com/pymmcore-plus/pymmcore-widgets/pull/414

Issue #440, opened 2025-02-21
bug: launching `mmcore mmstudio` from cli doesn't load plugins


tlambert03:
Presumably the method I used here to launch micro-manager is not properly using the java launcher and is therefore missing class paths:

https://github.com/pymmcore-plus/pymmcore-plus/blob/72e724c9cc5fd02482091311cfb11652e683ad38/src/pymmcore_plus/_cli.py#L133-L144

@marktsuchida, do you know how I should be invoking the ImageJ launcher programmatically so as to set up class paths correctly?

PR #459, opened 2025-04-09
Change to MM directory for `mmcore mmstudio`

marktsuchida:
Without this, especially on Windows, device adapters cannot find their dependency DLLs (among other possible problems due to MMStudio assuming its working directory to be the Micro-Manager folder).


src/pymmcore_plus/_cli.py
```diff
@@ -148,7 +149,8 @@ def mmstudio() -> None:  # pragma: no cover
         print("[magenta]run `mmcore install` to install a version of Micro-Manager")
         raise typer.Exit(1)
     cmd = ["open", "-a", str(app)] if PLATFORM == "Darwin" else [str(app)]
-    raise typer.Exit(subprocess.run(cmd).returncode)
+    with contextlib.chdir(str(mm)):
```
marktsuchida, line 152: Is there a better way than this to keep mypy from complaining that `mm` is `str | None`?

tlambert03:
Thanks!  Does this fix #440 ?

gselzer:
Works well for the use case I bothered @marktsuchida about! 😅

src/pymmcore_plus/_cli.py
tlambert03: This is essentially the same check as `if not app` above, so you could make it `if not mm or not app` and then mypy should narrow the type to str

marktsuchida: I though I tried that but I guess I didn't because it works now. Thanks.

marktsuchida:
> Thanks! Does this fix #440 ?

Yes, it does!

> Which was?

It was loading a device adapter (BH_DCC_DCU) that requires a vendor DLL (dcc64.dll) to be in the MM folder.

tlambert03:
sorry about the test fails, I thought I had fully reverted that...  will fix later

PR #461, opened 2025-04-16
fix: tensorstore - make sure metadata written

wl-stepp:
I ran into the situation where the tensorstore handler did not actually write the .zattrs file. I think it's because the async future is never waited for before the writer closes. Might not be a problem in most situations where the writer lives long enough, but might be a little more reliable like this.

Could also be added to the futures and awaited somewhere later, but was not sure when that would happen. But maybe on something like a __del__ or something like that...


tlambert03 (approved):
so sorry for the delay here!  Thank you!

PR #462, opened 2025-05-05
feat: add ValueBatcher pattern, for sharing control of a slow target value

tlambert03:
This is a GUI-agnostic helper class that is designed to fix cases like https://github.com/pymmcore-plus/pymmcore-widgets/issues/410, where we need to:

1. accumulate relative calls to a particular target value (as might be done when clicking on a move stage relative button multiple times before it has settled).
2. Possibly override the accumulated relative calls with an absolute setpoint, midstream.
3. Have multiple listeners be notified when it has reached a target and is idle (e.g. to snap an image)

It establishes a low level abstract class that generally implements the pattern in a device/value agnostic way, then adds a few concrete classes for StageDevice and XYStageDevice.

the primary public API is `CMMCorePlus.getDeviceBatcher`  (@marktsuchida, your opinions on API naming or other things very welcome here).

Below is an example of what this PR enables in pymmcore-widgets:

https://github.com/user-attachments/assets/75661234-4b48-40a3-9399-7fe674641df9

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,6 +115,8 @@ sources = ["src"]
 [tool.ruff]
 line-length = 88
 target-version = "py39"
+fix = true
+unsafe-fixes = true
 
 [tool.ruff.lint]
 pydocstyle = { convention = "numpy" }
--- a/src/pymmcore_plus/__init__.py
+++ b/src/pymmcore_plus/__init__.py
@@ -8,6 +8,7 @@ except PackageNotFoundError:  # pragma: no cover
     __version__ = "unknown"
 
 
+from ._batcher import AbstractValueBatcher, get_device_batcher
 from ._logger import configure_logging
 from ._util import find_micromanager, use_micromanager
 from .core import (
@@ -34,6 +35,7 @@ from .core.events import CMMCoreSignaler, PCoreSignaler
 from .mda._runner import GeneratorMDASequence
 
 __all__ = [
+    "AbstractValueBatcher",
     "ActionType",
     "CFGCommand",
     "CFGGroup",
@@ -59,5 +61,6 @@ __all__ = [
     "__version__",
     "configure_logging",
     "find_micromanager",
+    "get_device_batcher",
     "use_micromanager",
 ]
--- a/src/pymmcore_plus/_batcher.py
+++ b/src/pymmcore_plus/_batcher.py
@@ -1,0 +1,282 @@
+"""Batch setX calls to a device."""
+
+from __future__ import annotations
+
+import abc
+from abc import ABC, abstractmethod
+from collections.abc import Sequence
+from typing import Any, Generic, Literal, TypeVar
+
+import psygnal
+from typing_extensions import TypeAlias
+
+from pymmcore_plus.core._constants import DeviceType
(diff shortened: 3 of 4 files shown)


marktsuchida:
Wonderful to make this a general concept!

> the primary public API is `CMMCorePlus.getDeviceBatcher` (@marktsuchida, your opinions on API naming or other things very welcome here).

- It might be applicable to certain properties, at least for specific applications, so it might be good to avoid associating it with whole devices. Also I can think of it being applied to camera exposure (adjusting exposure for a slow camera) and gain (not currently standardized but could be) simultaneously. So having methods like `getPositionBatcher` on a Z stage object or `getPropertyBatcher` for numeric properties might be one way.

- As for the "batcher" part, the only alternative that comes to my mind is "jogger" or "jog controller" -- since this is primarily used for a jogging UI. That term seems to be fairly common in the motion control industry, for the push-to-move-fixed-increment functionality (and for press-and-hold-to-move-at-constant-velocity).

tlambert03:
> It might be applicable to certain properties, at least for specific applications, so it might be good to avoid associating it with whole devices. 

yeah, I was definitely imagining this could extend to properties, but stopped short of implementing it just yet.  Indeed there's a semantics challenge here.  In my head, I was associating a "device" batcher with "the one primary value that we associate with that device"... (such as a position with a stage, or a state with a state device) but that's really quite implicit, and some devices might have multiple values you want to batch control.    So, let's come up with the ideal future-proof API...

> So having methods like getPositionBatcher on a Z stage object or getPropertyBatcher for numeric properties might be one way.

so then, does that imply

```python
class CMMCorePlus:
    def getPositionBatcher(self, device_label: str): ...
    def getPropertyBatcher(self, device_label: str, property: str): ...
```

?

tlambert03:
> As for the "batcher" part, the only alternative that comes to my mind is "jogger" or "jog controller" -- since this is primarily used for a jogging UI. That term seems to be fairly common in the motion control industry, for the push-to-move-fixed-increment functionality (and for press-and-hold-to-move-at-constant-velocity).

but also a bit movement specific... i think i prefer `ValueBatcher` so far.  But how about `TargetAggregator` or `ValueAccumultor`?

tlambert03:
> I was thinking of only being able to access the batchers from the OO device object 

yeah, that's fine too!  It's probably better that I stop extending the top level API :)  I'll likely merge #437 soon then, and add it to that.

> `IncrementalChangeBatcher`

that's probably my favorite.. how about just `ChangeBatcher`?

gselzer:
For my 2 cents I like `Accumulator` or `Aggregator` over `Batcher` - using "batch" gives the impression that each atomic/relative movement is being retained in its own object, but really all of these small movements are being unified into one larger movement - I particularly think aggregate is a nice word to describe the diagonal movement that you get in your gif

tlambert03:
ok, sounds good.  `ChangeAccumulator` it is!  going once... going twice...

marktsuchida:
> ok, sounds good. `ChangeAccumulator` it is! going once... going twice...

I think that's a good one. If in practice it is prefixed with a value name (`PositionChangeAccumulator`), it reads pretty nicely.

tlambert03:
going in!  Thanks for the discussion all.  will release a new version and use over in widgets

PR #467, opened 2025-05-16
fix: dangling callback during core deletion, and disallow `registerCallback`

tlambert03:
This fixes a latent bug that was exposed in pymmcore `11.5.1.73.0`:

If a `pymmcore.MMEventCallback` is registered with the core, and a pointer is not retained, then a segfault can occur when the c++ core destructor is called:

```python
core = pymmcore.CMMCore()
core.registerCallback(pymmcore.MMEventCallback())
del core  # <- boom
```

This PR:
1. makes sure to call `super().registerCallback(None)` in the `__del__` method, before calling `core.reset()`
2. makes it an error to call `registerCallback` at all... since it would break everything.  Users are instructed to connect to `core.events` instead 

also uses and closes #466

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
         if: runner.os != 'Linux' && steps.cache-mm-build.outputs.cache-hit != 'true'
         run: uv run mmcore install
 
-      - run: uv run coverage run -p -m pytest -v --color=yes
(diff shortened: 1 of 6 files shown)


marktsuchida:
Looks good, thanks!

tlambert03:
the test error is real and I can reproduce.    it's specifically caused by the line `    pymmcore_plus._pymmcore.CMMCore.registerCallback(core, core._callback_relay)` ... but that's very confusing

PR #468, opened 2025-05-30
feat: add Unicore camera implementation

### LINKED ISSUE #465: Feat: add pure python Camera device ABC for unicore
@hinderling  mentioned he would like to use it in https://github.com/pymmcore-plus/pymmcore-plus/pull/407#issuecomment-2858988450

tlambert03:
this is a start on adding Camera devices to unicore.  closes #465 

@hinderling, it should already be working for a basic "snap" and/or "stream" if you want to try to check out this branch and implement a device adapter.  It should look roughly like this device in the test: 

https://github.com/tlambert03/pymmcore-plus/blob/fac9f321c32c7007efe1fd8974a9ca56e4172ec3/tests/unicore/test_camera.py#L21-L78

(and for a quick test, you don't need to use threading as I did there)


tlambert03:
Heres an example of the minimal camera device adapter:

```python
from collections.abc import Iterator, Mapping
from typing import Callable

import numpy as np

from pymmcore_plus.experimental.unicore import Camera
from pymmcore_plus.experimental.unicore.core._unicore import UniMMCore


class MyCamera(Camera):
    _exposure: float = 10.0

    def get_exposure(self) -> float:
        return self._exposure

    def set_exposure(self, exposure: float) -> None:
        self._exposure = exposure

    def shape(self) -> tuple[int, int]:
        return (480, 640)

    def dtype(self) -> np.typing.DTypeLike:
        """Return the data type of the image buffer."""
        return np.uint8

    def start_sequence(
        self, n: int, get_buffer: Callable[[], np.ndarray]
    ) -> Iterator[Mapping]:
        """Start a sequence acquisition."""
        shape, dtype = self.shape(), self.dtype()
        for _ in range(n):
            # acquire the image
            image = np.empty(shape, dtype=dtype)
            # update buffer in-place
            get_buffer()[:] = image
            yield {}  # any metadata


core = UniMMCore()
core.loadPyDevice("Camera", MyCamera())
core.initializeDevice("Camera")
core.setCameraDevice("Camera")
core.setExposure(42)
core.snapImage()
core.getImage()
```

and... if you make a nifty little image generator, here's an example of using a python camera device adaptor with pymmcore-widgets stuff (no changes needed to widgets at all!)


https://github.com/user-attachments/assets/fc187fba-b7c7-4fe8-ae98-aa581eedeaef

tlambert03:
Awesome!

The slm device is still a stub wip https://github.com/pymmcore-plus/pymmcore-plus/pull/414

But perhaps having spent a moment learning the new general pattern, you might have thoughts on how to proceed

tlambert03:
I'm pretty happy with the state of things here and I'm going to merge this.

I added an example with a real hamamatsu camera and was able to achieve the full 30FPS full frame bandwidth of the USB interface I was working with (not terribly impressive, but that's all the camera could give)

I tested bandwidth in silicon just feeding numpy frames into the system and just under 10GB/s, so I think that we're starting with a relatively good pattern in terms of performance.  

I do very much want @marktsuchida's review on this, but he's away for a bit.  So let's merge and continue to play with it, and when he has a moment we can integrate his suggestions.

PR #471, opened 2025-06-13
fix: camera fixes

tlambert03:
some suggestions from @marktsuchida


tlambert03:
small performance regression to be expected, due to move to copy-buffer by default

PR #485, opened 2025-07-11
test: add reference cleanup tests

tlambert03:
This adds some tests against strong-reference holding for CMMCorePlus, after some chats with @marktsuchida


tlambert03:
Eventually found the hangs, which were due to the ThreadPool used in waitDevices in the Unicore variant, specifically during object deletion (because `reset()` is called).  I made it so that the final call to `waitDevices` is always synchronous during object deletion, and that seems to have fixed it

PR #525, opened 2025-12-12
Support for config groups with UniMMCore devices

hinderling:
Hi! It's been a while since working on this, so please apologize that my memory of the issue is a bit hazy!

We ran into some problems when trying to use config groups with UniMMCore devices; basically it's not possible to use UniMMCore devices in config groups, as the logic (define/get/set config) is implemented at the level of MMCore (which can only see devices registered in the c++ core and doesn't have any knowledge about python-side devices). This leads to `RuntimeError: No device with label "MyUniMMCoreDevice"` when trying to apply a config group with a UniMMCore device. We added an additional layer in UniMMCore to catch such errors, and fall back to python-side handling of config groups. You can see in the tests that this allows us to mix python and native devices in config groups. 

While this fixed our issue, I'm sure this is not the best way to implement this; but I thought I will make a pull request to open the discussion. Do you think this would be in scope to add support for? 

Tagging @ddd42-star who first ran into the issue. Also, just letting you know that we're having a blast building virtual "fake" microscopes with UniMMCore devices!! Super useful for us.

### DIFF
--- a/src/pymmcore_plus/experimental/unicore/core/_unicore.py
+++ b/src/pymmcore_plus/experimental/unicore/core/_unicore.py
@@ -7,8 +7,8 @@ from collections.abc import Iterable, Iterator, MutableMapping, Sequence
 from contextlib import suppress
 from datetime import datetime
 from itertools import count
-from time import perf_counter_ns
 from pathlib import Path
+from time import perf_counter_ns
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -17,8 +17,6 @@ from typing import (
     TypeVar,
     cast,
     overload,
-    Iterable,
-    Union
 )
 
 import numpy as np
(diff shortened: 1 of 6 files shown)


tlambert03:
thanks @hinderling!  this is definitely in scope.  Indeed, all of the core-level functions that deal with "any" device do ultimately need to be rewritten to support pyDevices.  Thanks so much for getting this going.  I'll try to have a look at the implementation soon

tlambert03:
ok, I had a chance to start digging a bit.  There's a lot i still need to look through, but I can tell already that I'll want to break this into at least 2, maybe 3 or more PRs.  From the title, I assumed this was about config groups, but I can see it's much broader, including loading a system configuration `.cfg` file, and that has much bigger implications (though obviously, *is* a critical feature if this unicore is to be a long-lasting thing).  It would be the first time pymmcore-plus had introduced a feature that was strictly incompatible with switching back over to java micromanager with the same config file.  I do recognize that unicore devices themselves represent that shift, but I'd like to think very carefully about how an actual config file format would look like: it's important to me that we do it gracefully... Ideally in a way where the config could still be used in MMStudio, just without the pydevices (not a full on failure).  So, that part definitely needs to be separate.

If possible, I also like to consider cache warming for State/Label devices in initializeDevice/initializeAllDevices separately (and understand a bit more what you ran into that prompted it).

**I'd like this PR to *just* be able supporting config groups.  So, basically, all of the methods under `Config Group Methods`: `defineConfigGroup`, `renameConfigGroup`, etc...**

Can you create a new branch and PR just for programmatic control of config groups (leave this one open for discussion)?  (I know it's not yet the complete feature set you need, but that's step one).  If you want, I'm happy to branch your work into different branches, retaining your authorship, and prune it down.  Let me know...

dario-bassi:
Hi @tlambert03 , thank you again for your work. During our implementation of the virtual microscope, @hinderling and I decided to also enable the functionality to use MDA acquisition. When I tried to use the function as it is right now, I got the error from the function **device_info** (in mda/metadata/functions.py), where it could not access the parentLabel from the core. As a temporary solution, I added in  _unicore.py these two functions that solved the issue for our VM, and we were able to use the MDA acquisition for different channels. When @hinderling and I created the original PR #525, in the end, I decided not to include my workaround for this specific issue, mainly because I wasn't able to implement a reliable solution for our python devices. Below, I attach what I did:

    def getParentLabel(
        self, peripheralLabel: DeviceLabel | str
    ) -> str:
        return ""
   
    def setParentLabel(
        self, deviceLabel: DeviceLabel | str, parentHubLabel: DeviceLabel | str
    ) -> None:
        pass     

Do you think there is a better solution to this? I know that the MDA was developed for a physical microscope, but for us would be great if we could also use this feature.

tlambert03:
checking back in here...  a number of bits that were being done here were broken into smaller PRs:

- #525 
- #529
- #530
- #531
- #532 
- #534
- #540

@dario-bassi and @hinderling, when you have a moment, would you mind checking whether we've hit all the stuff here and if this can be closed?  (and actually... if we haven't, let's close anyway and open a TODO issue)

PR #534, opened 2026-01-08
feat: add missing propertyChanged signal for setProperty/setState/setStateLabel functions

dario-bassi:
Hey,
This is a continuation of #525 . I have noticed that for the UniMMCore core, the signal propertyChanged isn't emitted for the functions setProperty, setState, and SetStateLabel.

Furthermore, I think it would be helpful to mention in the documentation of the devices that the stagePositionChanged, XYStagePositionChanged, exposureChanged and SLMExposureChanged signals need to be added in the implementation of the Python device.

@hinderling and I managed to use our virtual microscope with napari-micromanager, with the added signals.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,13 +5,13 @@ ci:
 
 repos:
   - repo: https://github.com/adhtruong/mirrors-typos
-    rev: v1.40.0
+    rev: v1.41.0
     hooks:
       - id: typos
         args: [--force-exclude]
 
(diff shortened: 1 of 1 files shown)


tlambert03 (commented):
thanks @dario-bassi,  and thanks for including tests!  I'm going to suggest a big simplification though, and that is using unittests `Mock` (which is super handy when you want to make sure something has been called) like this.  Please have a look and merge the suggestions if you agree

dario-bassi:
Thanks @tlambert03 . I agree, it's much cleaner

PR #535, opened 2026-01-09
feat: Set `position_key` in `_5DWriterBase` to `event.pos_name` if available

ieivanov:
When writing data positions were previously labeled `f"{POS_PREFIX}{position_index}"` and metadata on the position name, if defined was lost. This PR fetches the position name from the MDA event and passes it to the writer.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,10 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "quarterly"
     commit-message:
       prefix: "ci(dependabot):"
+    groups:
+      actions:
+        patterns:
+          - "*"
--- a/src/pymmcore_plus/experimental/unicore/core/_config.py
+++ b/src/pymmcore_plus/experimental/unicore/core/_config.py
@@ -1,0 +1,706 @@
+"""Config file loading and saving for UniMMCore.
+
+This module provides Python-owned config file loading/saving that supports
+both C++ and Python devices. Python device lines use a `#py ` prefix so they
+are treated as comments by upstream C++/pymmcore implementations.
+
+Format example:
+    # C++ devices
+    Device,Camera,DemoCamera,DCam
+    Property,Core,Initialize,1
+
+    # Python devices (hidden from upstream via comment prefix)
+    #py pyDevice,PyCamera,mypackage.cameras,MyCameraClass
+    #py Property,PyCamera,Exposure,50.0
+    #py Property,Core,Camera,PyCamera
+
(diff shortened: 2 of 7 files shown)


tlambert03:
thanks @ieivanov!

>[!note]
> all of these handlers are slated for complete replacement with stuff in [ome-writers](https://github.com/pymmcore-plus/ome-writers) soon (which you may already know from working with @ jeskesen?).  I'm sorry that we don't immediately have a more stable/formalized writing path for you to adopt.  it's a high priority, but a tricky topic to get "just right".


 I'm definitely open to this change.  but could use a little bit more context: are you primarily using the Zarr subclass, the tiff subclass, or both?  
If zarr, then it's worth mentioning that the pattern we're aiming for in the ome-writers standardization would currently be the transitional bioformats2raw.layout spec, which is something like:

```
root
    ├── zarr.json             # Contains "bioformats2raw.layout" metadata
    ├── OME                   # Special group for containing OME metadata
    │   ├── zarr.json         # Contains "series" metadata
    │   └── METADATA.ome.xml  # OME-XML file stored within the Zarr fileset
    ├── 0                     # First image in the collection
    ├── 1                     # Second image in the collection
    └── ...
```

we *could* use position names instead of 0, 1, etc... there... but I almost feel like that metadata should be in the `zarr.json` somewhere, rather than the directory hierarchy itself.  

In any case: I'm definitely open to anything, but would like to have a better sense for what your ultimate desired outcome is (in terms of directory structure and/or metadata storage patterns), and whether it's tiff/zarr or both

ieivanov:
Hi @tlambert03! 

Yes, I am working with Justin on this, and I know it's a very dynamic issue.

For now I'm looking for a writer which captures basic acquisition metadata (e.g. channel names, z step size, etc), including position names, and writes data in a chunked format - something larger than single XY frames. Ideally we'd write data in zarr format, but we are working on a short turnaround project, so for now tif is also an option.

So far I've only tested the writers available here, but I'll try the [ome-writers](https://github.com/pymmcore-plus/ome-writers) next. With `OMEZarrWriter` and `TensorStoreHandler` I couldn't quite get the chunking to work, so I've chosen the `OMETiffWriter` as a fallback, as it writes all data from one position in a single tif file. As far as I can tell, the `OMETiffWriter` doesn't capture the position name, hence this PR (aside: I also found `metadata/_ome.py` but I'm not sure how it's used).

I'll test the ome-writers next, and I welcome any advice from you too! Thanks in advance.

ieivanov:
>so, this works for me! Normally, I'd say let's add a test to prevent regressions, but it's small and possible temporary :) so I'll leave it up to you. Good to merge?

Great, let's merge this PR then! Thanks for your input.

PR #538, opened 2026-01-29
fix: fix missing min_start_time on SequencedEvent

### LINKED ISSUE #537: Sequenced events do not respect the interval in the time loop.
I am not 100% sure this isn't user-error on my part, but I think this is the problem we're seeing here:  czbiohub-sf/shrimPy/issues/222

## Expected Behavior: Normal MDASequence Event Generation

When you create an MDASequence with time intervals:

```python
from useq import MDASequence
from datetime import timedelta

sequence = MDASequence(
    time_plan={"interval": timedelta(seconds=2), "loops": 3},
    z_plan={"range": 4, "step": 1},  # 5 z-slices
)
```

useq generates events with **different `min_start_time` values per timepoint**:

```
Timepoint 0 events:
  - Event(t=0, z=0, min_start_time=0.0)
  - Event(t=0, z=1, min_start_time=0.0)
  - Event(t=0, z=2, min_start_time=0.0)
  - Event(t=0, z=3, min_start_time=0.0)
  - Event(t=0, z=4, min_start_time=0.0)

Timepoint 1 events:
  - Event(t=1, z=0, min_start_time=2.0)  ← 2 seconds from sequence start
  - Event(t=1, z=1, min_start_time=2.0)
  - Event(t=1, z=2, min_start_time=2.0)
  - Event(t=1, z=3, min_start_time=2.0)
  - Event(t=1, z=4, min_start_time=2.0)

Timepoint 2 events:
  - Event(t=2, z=0, min_start_time=4.0)  ← 4 seconds from sequence start
  - Event(t=2, z=1, min_start_time=4.0)
  - Event(t=2, z=2, min_start_time=4.0)
  - Event(t=2, z=3, min_start_time=4.0)
  - Event(t=2, z=4, min_start_time=4.0)
```

### What happens in `_sequencing.py` can_extend() with DIFFERENT min_start_times:

When processing these events:

1. **Timepoint 0 batch starts**: `event_batch = [Event(t=0, z=0)]`

2. **Next event: Event(t=0, z=1)** - can we extend?
   ```python
   event.index.get("t") != e0.index.get("t")  # False (both t=0)
   and event.min_start_time != e0.min_start_time  # False (both 0.0)
   # Overall: False AND False = False → don't return False → CAN EXTEND ✓
   ```

3. **All t=0 events merge** into one SequencedEvent

4. **Next event: Event(t=1, z=0)** - can we extend the t=0 batch?
   ```python
   event.index.get("t") != e0.index.get("t")  # True (t=1 vs t=0)
   and event.min_start_time != e0.min_start_time  # True (2.0 vs 0.0)
   # Overall: True AND True = True → return False → CANNOT EXTEND ✓
   ```

5. **Result**: Correctly separates timepoints! ✓

---

## Problem Scenario: Hardware Sequencing with SAME min_start_time

In hardware sequencing mode (e.g., Mantis with TriggerScope), events are sometimes generated with `min_start_time=0.0` **relative to each timepoint**, not absolute from sequence start.

This happens because:
- Hardware sequences run immediately when triggered
- The engine manages timepoint timing separately from the hardware sequence
- Events within a timepoint all start at the same relative time (0.0)

Example event generation:

```
Timepoint 0 events:
  - Event(t=0, z=0, min_start_time=0.0)
  - Event(t=0, z=1, min_start_time=0.0)
  - Event(t=0, z=2, min_start_time=0.0)

Timepoint 1 events:
  - Event(t=1, z=0, min_start_time=0.0)  ← SAME as t=0!
  - Event(t=1, z=1, min_start_time=0.0)
  - Event(t=1, z=2, min_start_time=0.0)

Timepoint 2 events:
  - Event(t=2, z=0, min_start_time=0.0)  ← SAME as t=0!
  - Event(t=2, z=1, min_start_time=0.0)
  - Event(t=2, z=2, min_start_time=0.0)
```

### What happens in `_sequencing.py` can_extend() with SAME min_start_times:

1. **Timepoint 0 batch starts**: `event_batch = [Event(t=0, z=0)]`

2. **Next event: Event(t=0, z=1)** - can we extend?
   ```python
   event.index.get("t") != e0.index.get("t")  # False (both t=0)
   and event.min_start_time != e0.min_start_time  # False (both 0.0)
   # Overall: False AND False = False → don't return False → CAN EXTEND ✓
   ```

3. **All t=0 events merge** into batch

4. **Next event: Event(t=1, z=0)** - can we extend the t=0 batch?
   ```python
   event.index.get("t") != e0.index.get("t")  # True (t=1 vs t=0)
   and event.min_start_time != e0.min_start_time  # False (both 0.0) ← BUG!
   # Overall: True AND False = False → don't return False → CAN EXTEND ✗✗✗
   ```

5. **BUG**: t=1 events are added to the t=0 batch!

6. **Next event: Event(t=2, z=0)** - can we extend the t=0+t=1 batch?
   ```python
   event.index.get("t") != e0.index.get("t")  # True (t=2 vs t=0)
   and event.min_start_time != e0.min_start_time  # False (both 0.0) ← BUG!
   # Overall: True AND False = False → don't return False → CAN EXTEND ✗✗✗
   ```

7. **Result**: ALL timepoints merged into ONE SequencedEvent! ✗

---

## The Bug in `_sequencing.py` Lines 231-236

```python
# If it's a new timepoint, and they have a different start time
# we don't (yet) support sequencing.
if (
    event.index.get("t") != e0.index.get("t")
    and event.min_start_time != e0.min_start_time
):
    return False
```

### Truth Table Analysis:

| Different timepoint? | Different min_start_time? | Condition Result | can_extend returns? | Merges? |
|---------------------|--------------------------|-----------------|-------------------|---------|
| False (same t)      | False (same time)        | False           | True (continues)   | ✓ YES   |
| False (same t)      | True (diff time)         | False           | True (continues)   | ✓ YES   |
| True (diff t)       | False (same time)        | **False**       | **True** (continues) | **✗ YES (BUG!)** |
| True (diff t)       | True (diff time)         | True            | False (rejects)    | ✓ NO    |

**Row 3 is the bug**: Different timepoints with the same min_start_time are allowed to merge!

---

## The Fix

Change `and` to `or`:

```python
if (
    event.index.get("t") != e0.index.get("t")
    or event.min_start_time != e0.min_start_time  # ← Changed to OR
):
    return False
```

### Fixed Truth Table:

| Different timepoint? | Different min_start_time? | Condition Result | can_extend returns? | Merges? |
|---------------------|--------------------------|-----------------|-------------------|---------|
| False (same t)      | False (same time)        | False           | True (continues)   | ✓ YES   |
| False (same t)      | True (diff time)         | True            | False (rejects)    | ✓ NO    |
| True (diff t)       | False (same time)        | **True**        | **False** (rejects) | **✓ NO (FIXED!)** |
| True (diff t)       | True (diff time)         | True            | False (rejects)    | ✓ NO    |

Now different timepoints are **always** rejected, regardless of min_start_time!

---

## Impact on Time Intervals

When the bug causes all timepoints to merge into one SequencedEvent:

1. **MDARunner._wait_until_event()** is called **once** before the combined SequencedEvent
   - Waits until `min_start_time=0.0` (immediately)
   
2. **MDAEngine.exec_sequenced_event()** runs the entire hardware sequence
   - Acquires ALL z-slices for ALL timepoints back-to-back
   - No waiting between timepoints
   
3. **Result**: A 10-timepoint × 5-minute-interval acquisition (expected 45 minutes) completes in ~30 seconds

### With the fix:

1. **3 separate SequencedEvents** are created (one per timepoint)

2. **MDARunner waits before each timepoint**:
   - Wait until min_start_time=0.0 → run timepoint 0
   - Wait until min_start_time=2.0 → run timepoint 1  
   - Wait until min_start_time=4.0 → run timepoint 2

3. **Result**: Time intervals are respected! ✓

tlambert03:
should fix #537 

so sorry @jeskesen that it took me so long to understand.  the title itself is quite clear... the original description and the lack of a changing `min_start_time` in the problem scenario example got me hung up on the wrong details.

will add a test for this tomorrow, but if you want to try this and let me know if it solves your issue, that would be great confirmation

### DIFF
--- a/tests/test_sequencing.py
+++ b/tests/test_sequencing.py
@@ -6,7 +6,6 @@ from math import prod
 from typing import cast
 from unittest.mock import MagicMock, call
 
-import numpy as np
 import pytest
 import useq
 
@@ -184,26 +183,19 @@ def test_sequence_actions(core: CMMCorePlus) -> None:
     assert len(events) == EXPECTED_SEQUENCES
 
 
-def test_sequencing_respects_min_interval(core: CMMCorePlus) -> None:
+@pytest.mark.parametrize("interval", [0, 1])
+def test_sequencing_respects_min_interval(core: CMMCorePlus, interval: float) -> None:
     # Create sequence and generate events
-    interval = 0.5
     seq = useq.MDASequence(
         z_plan={"range": 1.0, "step": 1.0},
-        time_plan={"interval": interval, "loops": 2},
+        time_plan={"interval": interval, "loops": 3},
     )
 
     core.setExposure(1)
     core.setProperty("Z", "UseSequences", "Yes")
-
-    timestamps = []
-    core.mda.events.frameReady.connect(
-        lambda *a: timestamps.append(a[2]["runner_time_ms"])
-    )
-
     core.mda.engine.use_hardware_sequencing = True
-    core.mda.run(seq)
-
-    # ensure that intervals between frames are about interval (+/- 100 ms)
-    # increase atol if test is flaky on CI
-    intervals = np.diff(timestamps[::2])
-    np.testing.assert_allclose(intervals, interval * 1000, atol=100)
+    merged = list(iter_sequenced_events(core, seq))
+    assert len(list(seq)) == 6
+    assert len(merged) == 1 if interval == 0 else 3
+    expected_interval = [0, 1 * interval, 2 * interval] if interval else [0]
+    assert [e.min_start_time for e in merged] == expected_interval


tests/test_sequencing.py
```diff
@@ -181,3 +182,28 @@ def test_sequence_actions(core: CMMCorePlus) -> None:
     engine.use_hardware_sequencing = True
     events = list(engine.event_iterator(mda))
     assert len(events) == EXPECTED_SEQUENCES
+
+
+def test_sequencing_respects_min_interval(core: CMMCorePlus) -> None:
+    # Create sequence and generate events
+    interval = 0.5
```
jeskesen: would it also make sense to test that if an interval is 0, it all gets merged into a single event?

I know it was the use case I was avoiding, but it demonstrates proper sequencing over time... when requested. 

This test may already exist, but its tough to see from the perspective of a PR.

tlambert03:
> I'll have @ieivanov confirm it fixes the problem on our microscope.

thanks!  there's probably no need here... I'm 99% confident this will solve it, and it takes time to set this sort of test up.

PR #540, opened 2026-02-04
feat: getNumberOfCameraChannels to use MDA with python camera

dario-bassi:
Hey,
This is a continuation of #525 . We wanted to use MDA Acquisition with our Python camera device, but unfortunately, we couldn't because during MDA acquisition, the function getNumberOfCameraChannels is called. I wasn't sure if there was a reason why the function wasn't implemented for python devices or not, anyway I propose a simple implementation.

I wanted to leave the option of not adding this function as mandatory, but pre-commit did not accept my changes.

### DIFF
--- a/tests/00_unicore/test_camera.py
+++ b/tests/00_unicore/test_camera.py
@@ -68,10 +68,6 @@ class MyCamera(CameraDevice):
             buffer[:] = FRAME
             yield {"random_key": f"value_{i}"}  # Example metadata, can be anything.
 
-    def get_number_of_camera_channels(self) -> int:
-        """Returns the number of components the default camera is returning."""
-        return 1
-
 
 class SequenceableCamera(MyCamera):
     """Camera device that supports exposure sequencing."""


tlambert03:
thanks @dario-bassi ... let's definitely fix whatever underlying issues there are when using a python camera in MDA.

the reason this wasn't added to the python side is that it's actually a super confusing API.  It does *not* mean what you would think it would mean, and what your docstring (understandably) says it is:

> """Returns the number of components the default camera is returning."""

there you would expect an RGB camera to return 3... but that's actually `getNumberOfComponents`.

The origin is a little confusing to me, but I think [it was introduced](https://github.com/micro-manager/mmCoreAndDevices/commit/172776072cf5e98ee7d21687bd5eb26fa6b78d68) to support hardware multi-detector devices (like TwoPhoton/BitFlow) - single sensor with multiple independent detectors.  And then [later](https://github.com/micro-manager/mmCoreAndDevices/commit/291dae85fc493f8e03e27224c49d31315b88f31a) (and now more commonly) it was used as a bit of a hack to create "virtual" devices (like MultiCamera) that combine 1 or more physical devices... where `getNumberOfCameraChannels` then means the number of physical devices that the virtual device is controlling.

Suffice it to say that those are both unusual cases: I didn't want to expose that confusion in our clean new device ABC, and even if we did want to support something like that later, I think i'd want to do it in a different way.

**SO**

I think for your case here,  let's get rid of all the changes to the camera device itself, and simply replace that `NotImplementedError` inside of `getNumberOfCameraChannels` with `return 1`

tlambert03:
thanks @dario-bassi !  not sure what's going on with those test failures but obviously unrelated.

PR #541, opened 2026-02-04
fix: fix stage callback tests

tlambert03:
Commit [9bcc3b161](https://github.com/micro-manager/mmCoreAndDevices/commit/9bcc3b1612be014ca0a2ccde107b443879f11971) in the mmCoreAndDevices repo removed the
  `OnXYStagePositionChanged` callback from the `CXYStageBase::SetPositionUm` base class in MMDevice/DeviceBase.h.

  Before (old device adapters)

  When setXYPosition(1.0, 1.0) was called, the base class SetPositionUm did:
```cpp
  int ret = this->SetPositionSteps(xSteps, ySteps);
  if (ret == DEVICE_OK) {
      xPos_ = x_um;          // stores 1.0 (the REQUESTED position)
      yPos_ = y_um;
      this->OnXYStagePositionChanged(xPos_, yPos_);  // fires with (1.0, 1.0)
  }
```
  So the callback was emitted with the exact requested position (1.0, 1.0).

  After (new device adapters)

  The base class now just does:
```cpp
  int ret = this->SetPositionSteps(xSteps, ySteps);
  if (ret == DEVICE_OK) {
      xPos_ = x_um;
      yPos_ = y_um;
      // NO OnXYStagePositionChanged callback!
  }
```

The only callback now comes from the `DemoXYStage::SetPositionSteps` in the device adapter itself, which uses the quantized
  position. The XY stage step size is 0.015 um, so:
  - nint(1.0 / 0.015) = 67 steps
  - 67 * 0.015 = 1.005 um

Tests were failing because the actual callback is now (1.005, 1.005) instead of (1.0, 1.0).  So we just round now

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,13 +5,13 @@ ci:
 
 repos:
   - repo: https://github.com/adhtruong/mirrors-typos
-    rev: v1.41.0
+    rev: v1.42.3
     hooks:
       - id: typos
         args: [--force-exclude]
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.10
+    rev: v0.14.14
     hooks:
       - id: ruff-check
         args: [--fix, --unsafe-fixes]


tlambert03:
remaining failures are in widgets

PR #545, opened 2026-02-13
feat: add runner handlers that do not rely on signals

fdrgsp:
@tlambert03 this is the PR we talked about. It is a WIP to add a `MDARunner`-managed writer based on `ome-writers` that does not rely on signals (and, if the writer fails, it will block the acquisition).

For now, the existing `output` (signal-based) logic is unchanged, this PR only adds a new `writer` argument to `core.mda.run`.

There are two new objects:
- **`OMERunnerHandler`**:  prepares the `ome-writers.OMEStream`, writes  frames (sync or async), and cleans up at the end.
- **`StreamSettings`**: configuration object passed to `OMERunnerHandler`. Currently subclasses `ome-writers.AcquisitionSettings` with `dimensions`, `dtype` and `plate` made optional since they can be derived from `MDASequence` and metadata at runtime (maybe better as a standalone internal dataclass since I had to call private validation methods?). It also includes an `asynchronous` flag that if `True` writes frames on a separate thread to avoid blocking the runner (not sure this is the best approach, but I think it works).

## Usage 

As in *examples/runner_writer.py*:

```python

# ------------------------------------------------------------------
# simply pass the path WITH extension
# ------------------------------------------------------------------
mmc.run_mda(sequence, writer="example.ome.zarr")

# ------------------------------------------------------------------
# use StreamSettings for more control
# ------------------------------------------------------------------
stream_settings = StreamSettings(root_path="example.ome.zarr", overwrite=True, asynchronous=True) 
mmc.run_mda(sequence, writer=stream_settings)

# ------------------------------------------------------------------
# manually create the handler and pass it to run_mda
# ------------------------------------------------------------------
stream_settings = StreamSettings(root_path="example.ome.zarr", overwrite=True, asynchronous=True)
handler = OMERunnerHandler(stream_settings)
mmc.run_mda(sequence, writer=handler)
```

We can still have more than one writer:
```py
# ------------------------------------------------------------------
# for multiple writers, pass a list of paths, settings, or handlers
# ------------------------------------------------------------------
mmc.run_mda(sequence, writer=["example.ome.zarr", "example1.ome.zarr"])
```

And once https://github.com/pymmcore-plus/ome-writers/pull/95 lands, I think we could expose the data view via a `OMERunnerHandler` property:

```py
@property
def view(self) -> omew.AcquisitionView | None:
    """The AcquisitionView object for viewing the stream."""
    if self._stream is None:
        return None
    return omew.AcquisitionView.from_stream(self._stream)
```

NOTE: test `macos-latest py3.10 PyQt6 lowest-direct` fails because we need the updated version of `yaozarrs`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,7 @@ docs = [
     "mkdocs-typer ==0.0.3",
 ]
 test = [
-    "pymmcore-plus[simulate]",
+    "pymmcore-plus[io,simulate]",
     "msgspec >= 0.19",
     "msgpack >=1",
     "pytest-cov >=7",


src/pymmcore_plus/mda/handlers/_runner_handler.py
```diff
@@ -0,0 +1,338 @@
...
+        `asynchronous` is True. Default is 100.
+    """
+
+    dimensions: DimensionList | None = None
+    dtype: DTypeStr | None = None
+    plate: omew.Plate | None = None
+    asynchronous: bool = True
+    queue_maxsize: int = 100
+
+    def model_post_init(self, __context: object) -> None:
```
tlambert03: ah very interesting! :) 

but... this isn't the right way to fix this.  The real bug is actually in our example that ends with a call to `run_mda()`  (instead of `mda.run()`) and doesn't block or wait for the thread to finish before exiting the script.  we shouldn't be showing that pattern.  Get rid of this bit that imports private stuff from ome_writers and just change the example (and also the run_mda.py example, to call `mda.run()`, or `run_mda(..., block=True)`

fdrgsp:
@tlambert03 I think I updated the code as you suggested (class naming can of course be changed):

- `core.mda.run()` has the same signature as before. I only added `BaseRunnerHandler` (a `Protocol` that implements `prepare`, `writeframe`, and `cleanup`) to the `SingleOutput` type alias.
- when `output` is passed as a path (or sequence of paths), we internally use `OMERunnerHandler` (via `OMERunnerHandlerGroup`) which is `MDARunner`-managed and does not rely on signals.
- for now I kept `StreamSettings` as the object used to create an `OMERunnerHandler` but it's a standalone `dataclass` and no longer a subclass of `ome_writers.AcquisitionSettings`.

Everything is kept private for now.

PR #550, opened 2026-02-19
feat: getROI/setROI/clearROI for UniMMCore Cameras

hinderling:
`getROI()`, `setROI()`, and `clearROI()` previously raised `NotImplementedError` for Python cameras, which broke MDA acquisition (the engine calls `getROI()` during `setup_sequence` to compute grid FOV sizes).

- Add `get_roi()`, `set_roi()`, `clear_roi()` to `CameraDevice` base class with sensible defaults (full frame from `shape()`)
- Wire up `UniMMCore.getROI/setROI/clearROI` to delegate to these methods, matching the existing pattern used by `getExposure`, `setExposure`, etc.
- Subclasses can override to support hardware ROI

Added tests to demo camera.


hinderling:
Thanks for all the comments!! I really like the idea of `SimpleCameraDevice`, this would 100% cover my use case, and hopefully not mislead people into using it with real hardware.

PR #553, opened 2026-02-20
fix: Teardown of sequenced events

ieivanov:
Sequenced events were not reset correctly before as the input to `teardown_event` was not the original sequenced event, but rather the last single event output by `exec_event`


tlambert03:
good catch. yeah, would be helpful to see a small test here to prevent regression

tlambert03:
test added

ieivanov:
Thank you!

PR #554, opened 2026-02-20
fix: stop acquisition thread in unloadAllDevices

hinderling:
Fixes a bug when loading a new .cfg in napari-micromanger, while live-mode is running. When `unloadAllDevices()` is called (e.g. before loading a new config), any in-flight sequence acquisition thread keeps running against the objects it holds, which can cause crashes or stale state. Stop and join the thread before clearing devices.


tlambert03:
thanks!

PR #557, opened 2026-02-23
feat: formalize MDARunner state machine, add status property

tlambert03:
this formalizes the state machine modeled by the MDARunner, and 

- adds a `status` method that summarizes current state (we may later deprecate things like "is_running" and "is_paused")
- simplifies a number of state branches, particularly in `_wait_until_event`
- adds simple state machine tests
- makes cancel/pause mutations threadsafe
- deprecates `toggle_pause()` and introduces `set_paused(bool)`

```mermaid
    stateDiagram-v2
        [*] --> IDLE
        IDLE --> PREPARING : run()
        PREPARING --> running : sequence ready
        state running {
            WAITING --> ACQUIRING : next event ready
            ACQUIRING --> WAITING : event done
            WAITING --> PAUSED : <code>toggle_pause()</code>
            PAUSED --> WAITING : <code>toggle_pause()</code>
            ACQUIRING --> PAUSED : <code>toggle_pause()</code></br>(after event done)
        }
        running --> FINISHING : <code>cancel()</code>
        running --> FINISHING : all events exhausted
        FINISHING --> IDLE : cleanup done
```

-----

side-effects of this PR:

- `sequenceCanceled` signal now fires after `teardown_sequence` instead of before.  (arguably more correct... i don't think it will affect any downstream guis)

also note:

- `toggle_pause()` during ACQUIRING emits signal immediately but defers actual state change  (this is how it's always been on main as well)

### DIFF
--- a/src/pymmcore_plus/mda/_runner.py
+++ b/src/pymmcore_plus/mda/_runner.py
@@ -13,6 +13,7 @@ from typing import TYPE_CHECKING, Any, cast
 from unittest.mock import MagicMock
 from weakref import WeakSet
 
+from typing_extensions import deprecated
 from useq import MDASequence
 
 from pymmcore_plus._logger import exceptions_logged, logger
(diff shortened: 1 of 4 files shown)


tlambert03:
> but is there a reason to prefer toggle_pause() instead of a set_paused(bool)

no, and I was thinking the same thing here.  the only reason is really more of a convenience for the GUI (easier to connect to button click) ... but that should not be the concern of pymmcore-plus, and the gui still has work to do anyway... so we should change that.

> if pause is turned on while ACQUIRING, and the transition to PAUSED only happens after the current event, does that mean the state remains unchanged...

 When toggle_pause() is called while ACQUIRING:

```python
# toggle_pause()
elif self._state == RunState.ACQUIRING:
    self._pause_requested = not self._pause_requested  # toggles the flag
    paused = self._pause_requested
# signal emitted outside lock
self._signals.sequencePauseToggled.emit(paused)
```

Then at the event boundary (after the current event finishes):

```python
  if self._pause_requested:
      self._pause_requested = False
      self._state = RunState.PAUSED

  if self._state != RunState.PAUSED:
      self._state = RunState.WAITING
```

So multiple toggles during ACQUIRING work correctly (an odd number of toggles will cause PAUSED at the event boundary... and an even number just return to the WAITING state).

but, again, I totally agree that toggle is the wrong thing to have here altogether... so I'll add set_paused, and deprecate toggle

PR #558, opened 2026-02-24
fix: redundant cpp calls

hinderling:
This should address: https://github.com/pymmcore-plus/napari-micromanager/pull/365, what claude suggested after a lot of back and forth, but definitely outside my understanding. Would be great if you can have a look @tlambert03 , if this is actually the issue or if this will lead to other problems.

**Summary:**
`CMMCorePlus.__del__` was calling `registerCallback(None)`, `reset()`, and `setPrimaryLogFile("")`, all C++ calls. But the C++ `~CMMCore()` destructor already does the exact same three things immediately after. This is both redundant and leads to the failed test in the PR above: when Python's GC triggers `__del__` during unrelated C++/Qt teardown (e.g. napari vispy layer removal), re-entering C++ causes an access violation crash on Windows.

**Fix:** 
Remove the C++ calls from `__del__`, keep only the `atexit.unregister()`. The C++ destructor handles everything.

### DIFF
--- a/src/pymmcore_plus/core/_mmcore_plus.py
+++ b/src/pymmcore_plus/core/_mmcore_plus.py
@@ -334,13 +334,19 @@ class CMMCorePlus(pymmcore.CMMCore):
         return f"<{type(self).__name__} at {hex(id(self))} with {ndevices} devices>"
 
     def __del__(self) -> None:
-        if hasattr(self, "_weak_clean"):
-            atexit.unregister(self._weak_clean)
-        # NOTE: we intentionally do NOT call registerCallback/reset/setPrimaryLogFile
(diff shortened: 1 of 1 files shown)


tlambert03:
for https://github.com/pymmcore-plus/napari-micromanager/pull/365 ... maybe change the pyproject for now to point to  this main branch here on github for now.  I don't want to immediately release a new version here cause some other things are in flight.  so, an official release of napari-micromanager might also have to wait for that

hinderling:
Yes, tests are also still failing here https://github.com/pymmcore-plus/napari-micromanager/pull/365... Maybe better to undo in the meantime?

PR #560, opened 2026-02-24
feat: Add eventFinished signal

ieivanov:
This PR adds an `eventFinished` signal which emits after `teardown_event`. Here is a summary of the sequence of events during an acquisition:
* `sequenceStarted`: Once, after `engine.setup_sequence()`
* `eventStarted`: Per event, before `setup_event()` + `exec_event()`
* `frameReady`: Per frame, inside `exec_event()`
* `eventFinished` **(new)**: Per event, after `teardown_event()`
* `sequenceFinished`: Once, after `teardown_sequence()`

This PR also removes the empty `mmCoreAndDevices` folder


ieivanov:
The tests are entirely Claude Code's idea, I'm not sure if they are appropriate

ieivanov:
Awesome, let me clean up the code I'm working on a bit and I'll share it with you

PR #562, opened 2026-02-25
fix: correctly mock PYMMCORE_DIV in tests

marktsuchida:
`pymmcore_plus.install.PYMMCORE_DIV` is set at import time, so patching `pymmcore_plus._pymmcore.version_info` doesn't work for these tests.


tlambert03 (approved):
ahh thank you

PR #572, opened 2026-02-27
fix: remove all CMMCore cleanup/deletion magic, improve unimmcore cleanup

### LINKED ISSUE #570: Meta-issue for segfaults on core cleanup
We'll still have some slightly messy stuff going on in the core cleanup, as it relates to the C++ level cleanup.   Just want to link all the relevant issues for now, and will summarize later:

- problem was described in https://github.com/micro-manager/mmCoreAndDevices/issues/594
- https://github.com/micro-manager/mmCoreAndDevices/pull/628 added `onSystemConfigurationLoaded` in `unloadAllDevices` 
- The version was bumped from 11.5.0 → 11.5.1 in commit https://github.com/micro-manager/mmCoreAndDevices/commit/f72bd4dd
      The destruction call chain becomes:
    
    ```C++
      ~CMMCore()
        → reset()
          → unloadAllDevices()   // ← NEW: now calls externalCallback_->onSystemConfigurationLoaded()
    ```
- In MMCore 11.5.0, unloadAllDevices() never touched `externalCallback_`, so destruction was safe even if the Python callback object had already been garbage collected.
- In MMCore 11.5.1, unloadAllDevices() now calls back into `externalCallback_`. Since `externalCallback_` is a raw pointer (CMMCore doesn't own it and never clears it), if the  Python-side MMEventCallback subclass has been GC'd, the SWIG director tries to invoke a method on  deallocated memory → segfault.
- https://github.com/micro-manager/pymmcore/commit/f9e27c3 patched the destructor to call registerCallback(nullptr) at the very start of `~CMMCore()`
- https://github.com/pymmcore-plus/pymmcore-plus/pull/467 fixed problems that now occurred on the python side
- https://github.com/pymmcore-plus/napari-micromanager/pull/365 is trying to do some jujitsu with core swapping at runtime (run failures like [runs/22372547896/job/64754945418](https://github.com/pymmcore-plus/napari-micromanager/actions/runs/22372547896/job/64754945418?pr=365))
- https://github.com/pymmcore-plus/pymmcore-plus/pull/558  was an attempt to fix it, but has brought back the segfaults
- https://github.com/pymmcore-plus/pymmcore-plus/pull/569  reverted 558

cc @hinderling

tlambert03:
closes #570 (or at least tries to)

CMMCorePlus now does **no special cleanup** on object deletion:
 
- get rid of the atexit `unloadAllDevices` that was originally added in https://github.com/pymmcore-plus/pymmcore-plus/pull/47 (but which was a no-op until @hinderling caught that and "fixed" it in https://github.com/pymmcore-plus/pymmcore-plus/pull/559  ... which actually seems to have exposed a problem, but only in benchmarks, when running under valgrind)
- get rid of any call to `registerCallback(None)` during destruction (it's called by the C++ code)
- Don't double call `reset()` etc...
- don't clear the primary log file...

### DIFF
--- a/src/pymmcore_plus/core/_mmcore_plus.py
+++ b/src/pymmcore_plus/core/_mmcore_plus.py
@@ -4,6 +4,7 @@ import os
 import re
 import time
 import warnings
+import weakref
 from collections import defaultdict
 from contextlib import contextmanager, suppress
 from datetime import datetime
@@ -180,7 +181,7 @@ def _blockSignal(obj: Any, signal: Any) -> Iterator[None]:
         obj.blockSignals(False)
 
 
-_instance = None
+_instance: weakref.ref[CMMCorePlus] | None = None
 
 
 class CMMCorePlus(pymmcore.CMMCore):
@@ -221,9 +222,10 @@ class CMMCorePlus(pymmcore.CMMCore):
 
         """
         global _instance
-        if _instance is None:
-            _instance = cls()
-        return _instance
+        inst = _instance and _instance()
+        if inst is None:
+            inst = cls()
+        return inst  # type: ignore[return-value]
 
     def __init__(self, mm_path: str | None = None, adapter_paths: Sequence[str] = ()):
         super().__init__()
@@ -241,8 +243,8 @@ class CMMCorePlus(pymmcore.CMMCore):
 
         # Set the first instance of this class as the global singleton
         global _instance
-        if _instance is None:
-            _instance = self
+        if _instance is None or _instance() is None:
+            _instance = weakref.ref(self)
 
         if hasattr(self, "enableFeature"):
             strict = True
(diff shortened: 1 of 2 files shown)


tlambert03:
ok, after struggling with this a lot.  I think we might be at around the best we can do.  there are still occasional segfaults that I'm told (by claude, repeatedly) that we probably can't fix without some TOCTOU fixes upstream in MMCore. 

changes

  **CMMCorePlus:**
  - `_instance` → weakref.ref (allows GC during normal execution)
  - `atexit` handler removed (C++ destructor handles it)
  - `__del__` reduced to just super().registerCallback(None) with suppress(Exception) 
  - Old heavy `__del__` (reset, setPrimaryLogFile) removed 
  - `core.__del__()` explicit call removed from test fixture 
  - CLI test mocks updated for weakref (lambda: mock) 

  **UniMMCore:**
  - `weakref.finalize` with extracted references 
  - `_cleanup_python_state` as static method 
  - `_reset_python()` factored out, called from both reset() and unloadAllDevices() 
  - Old reset() that mixed Python and C++ waiting removed 
  - Buffer size env var / CI detection (orthogonal)

PR #585, opened 2026-03-05
feat: Terminate stalled acquisitions

ieivanov:
This PR introduces a private `_sequenced_acq_timeout` which can be used to terminate stalled sequenced acquisitions due to a missed camera trigger purse or dropped frame. The default of `_sequenced_acq_timeout = np.inf` does not change the current behavior of `MDAEngine`. If `_sequenced_acq_timeout` is set to a non-zero value, the acquisition will be terminated when that deadline is exceeded. My current use patter looks something like:

```python
class MyMDAEngine(MDAEngine):

        # Call parent setup_event, sets channel and exposure time
        super().setup_event(event)

       framerate = 1000 / event.exposure  # approximate
       buffer = 2
       self._sequenced_acq_timeout = len(event.events) / framerate + buffer
```

This is a follow-up on our conversation in https://github.com/pymmcore-plus/pymmcore-plus/pull/578#issuecomment-3988040727

### DIFF
--- a/src/pymmcore_plus/model/_device.py
+++ b/src/pymmcore_plus/model/_device.py
@@ -195,7 +195,43 @@ class Device(CoreObject):
         self.device_type = core.getDeviceType(self.name)
         self._update_core_getters()
 
+        # Preserve the current adapter_name before the super call overwrites it.
+        # Some device adapters report a different name via core.getDeviceName()
+        # than the name required by core.loadDevice(). For example, the TSI
+        # adapter registers as "TSICam" for loading but getDeviceName() returns
+        # "TSI3Cam" after initialization. If the core-reported name is not in
+        # the list of available (loadable) device names for the adapter library,
+        # we keep the original adapter_name so that saved configs remain valid.
+        prev_adapter_name = self.adapter_name
+
         super().update_from_core(core, exclude=exclude, on_err=on_err)
+
+        if (
+            prev_adapter_name
+            and self.adapter_name != prev_adapter_name
+            and self.library
+        ):
+            try:
+                available = core.getAvailableDevices(self.library)
+            except RuntimeError:
+                available = ()
+            if available and self.adapter_name not in available:
+                import logging
+
+                logger = logging.getLogger(__name__)
+                logger.warning(
+                    "Device adapter %r reported device name %r via "
+                    "getDeviceName(), but that name is not in the loadable "
+                    "devices for library %r: %s. Keeping original adapter "
+                    "name %r to ensure saved configurations remain valid.",
+                    self.library,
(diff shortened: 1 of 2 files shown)


tlambert03 (approved):
this seems like a good feature to me.  @marktsuchida, any comments?

@ieivanov, I think the only question I have is: how would you intend to use this downstream?  I would not want you (or any external library subclassing MDAEngine) to access a private attribute of the super-class.  All private attributes should be considered internal implementation details that can change without deprecation warnings

tlambert03:
i think for now, unless you anticipate wanting to make it a computed property, just making the attribute name public is fine.  something like what we do for the others:

https://github.com/pymmcore-plus/pymmcore-plus/blob/079ae35076ed28e77cd9b149a86c5c5736c0a2d5/src/pymmcore_plus/mda/_engine.py#L104-L106

(as a side note, since you submitted https://github.com/pymmcore-plus/pymmcore-plus/pull/548, it's perhaps worth pointing out that I *do* intend for any non-private attributes to be user settable, even if they're not exposed in the `__init__` method).  So, if you want to just make it a public attribute that's fine, but also feel free to add it to the `__init__` if you prefer using it that way.

tlambert03:
> We should throw an error message that suggests the user increase these timeouts if they believe this happened in error.

sounds good.  I was actually going to ask you about this:   When should we fail outright, vs just fill in some missing frames and try again?  Do you definitely find cases where it's just a couple dropped frames as opposed to a missed first (or all) trigger?    

If it's worth "trying again", then we should just let it fall through to the code that yields `None` the appropriate amount of times, but if we're in a case where we know it's just "broken" and never going to work, then just raise a `RuntimeError()` with a message.  (and we'll get that error message to the user somehow)

ieivanov:
>When should we fail outright, vs just fill in some missing frames and try again? Do you definitely find cases where it's just a couple dropped frames as opposed to a missed first (or all) trigger?

Yeah, that's a tricky one. I'm currently working to resolve the one-in-a-million situation. We would often set up 2-3 day long acquisitions and would see 1-2 frames go missing. Just bad luck. In those cases, I think it's best to yield `None`, issue a warning and carry on.

We could add a counter that throws a `RuntimeError` when `self.max_num_missed_frames` is reached? That would make sense in case things have gone far too wrong or are just not set up correctly.

As for trying again - I think we should allow for cases where the user wants to retry a given event execution if it did not fully succeed the first time around. I think this should happen through a subclass that overrides one of the `MDAEngine` methods. I'm not sure what would be most ergonomic.

tlambert03:
thanks @marktsuchida!  I appreciate the look on this one

Issue #573, opened 2026-02-28
Clarifying singla emission threads


marktsuchida:
I'm not yet sure if this is a documentation issue or something more.

If I'm understanding correctly:

- Connecting with `@core.mda.events.frameReady.connect` results in events being emitted on the MDA runner thread
- Connecting with `run_mda(output=...)` results in events being relayed (via a queue) onto a dedicated thread, from which they are emitted onto the thread that called `run_mda()`
- Events coming from MMCore are emitted on whatever thread MMCore invoked the callback on

But:

- Qt signal [connections](https://doc.qt.io/qt-6/threads-qobject.html#signals-and-slots-across-threads), by default, automatically use a queue when the receiving object "has affinity" with a different thread
- psygnal emits on the same thread by default; an explicit thread can be specified

The problem is with events coming from MMCore because those can be either on a thread calling into MMCore or on a device-owned thread (even for the same event, such as `onPropertyChanged()`).

The "right" thing for MMCore to do is probably to internally queue all events and emit them asynchronously from a dedicated background thread. Aside from making the behavior consistent, this would prevent all kinds of problems that can arise from nested calls and callbacks (deadlocks and races). But this would mean that all `CMMCorePlus.events` events will be on a different thread, at least in the psygnal case (and I don't know how Qt behaves for Python receivers either) — even though, arguably, code that doesn't handle async events is already broken.

Is my description accurate? I could easily be missing something here, since I haven't dug into the weeds of Qt signals and slots.

As an aside, MMStudio handles this by relaying all events to the UI event dispatch thread. That may also largely work for Python Qt apps, but, of course, something needs to be done when Qt is not being used.

tlambert03:
thank you so much for digging into this!

The honest (and sadly obvious) answer is that **pymmcore-plus doesn't currently make a good guarantee about which thread events arrive on**.  And yeah: particularly with MDARunner signals, the thread affinity is even more variable.  The double psygnal/Qt signal system obviously complicates things (but, I still value that design, even if it's tricky: i don't want pymmcore-plus to have a Qt dependence, just a compatibility... though we *could* consider strictly using psygnal signals, which have generally worked find alongside real qt signals)

[...]

PR #588, opened 2026-03-07
fix: update _MMCallbackRelay for v12

tlambert03:
This PR makes changes to `_mmcore_plus.py` to solve issues that will arise with the [v12 MMCore changes](https://github.com/micro-manager/mmCoreAndDevices/pull/877).

This is related to (but does not solve #573)

problems:
- `setChannelGroup()` now emits `channelGroupChanged` _twice_ (once on main, once async)
- our old `_blockSignal` strategy breaks, it cannot catch async delivery — it's a race condition
- `snapImage()` emits `propertyChanged(shutter, "State", True)` with bool instead of str
- Event ordering changed: `configSet` now arrives before async `propertyChanged`

1. Removed `_blockSignal` function and its SignalInstance import — (no longer works, and no longer needed)
1. Restructured `_MMCallbackRelay` with three signal categories:
    - Relay-skip — CMMCorePlus method is sole emitter; relay drops C++ callback entirely (`onImageSnapped`, `onSequenceAcquisitionStopped`, etc...)
    - Relay-only — C++ callback is sole source; no manual emission exists (unchanged)
    - Custom relay — Needs dedup logic because both sources are legitimate (`onPropertyChanged`)
1. Value-aware dedup mechanism for `propertyChanged`:
  - Per-property queue of `(value, expiry)` tokens with TTL cleanup
  - register_property_emission() — registers a token when manually emitting
  - `onPropertyChanged()` — consumes matching token or relays through
  - `suppress_property()`/`unsuppress_property()` — wildcard suppression for v11 sync callbacks during `_property_change_emission_ensured`
1. Fixed `snapImage` — emits "1"/"0" strings instead of `True/False` bools, registers dedup tokens
1. Simplified `setChannelGroup` — no more `_blockSignal`, relay is in `_RELAY_SKIP`
1. Updated `setShutterOpen` and `setFocusDevice` — register dedup tokens before manual emission
1. Refactored `_property_change_emission_ensured` — uses suppress/unsuppress during yield (handles v11 sync),
 then registers value-aware tokens after (handles v12 async)

----

much of this is claude-assisted and I still need to do a careful sanity review.  Particularly with regards to the deduping of `propertyChanged`.  the need does generally make sense to me, but it's touching on a long history of issues

- https://forum.image.sc/t/micromanager-events-core-events-not-coming-through/53014
- https://github.com/micro-manager/mmCoreAndDevices/issues/25
- #38 
- #97 
- #587

it's worth "re-capping" here what the general problem is:

The fundamental issue is that MMCore's `setProperty()` does _not_ fire `onPropertyChanged`. It just calls `pDevice->SetProperty(propName, propValue)` and it's entirely up to the device adapter whether to call `OnPropertyChanged()`.  Which means the application has no guarantees about notifications, which are necessary for updating UI.

As Nico put it [on image.sc](https://forum.image.sc/t/micromanager-events-core-events-not-coming-through/53014): "Callbacks are mainly used to give devices the opportunity to signal back to the UI." ... But I'm still not sure that's a strictly accurate summary of the actual architecture, since there are also cases where MMCore.cpp itself directly fires callbacks (`onImageSnapped`, `onChannelGroupChanged`, `onSystemConfigurationLoaded`) and also `CoreCallback.cpp` fires callbacks on behalf of devices for various cases of property changes as well.  So at best, MMCore is internally inconsistent with itself: It fires `onChannelGroupChanged` from `setChannelGroup()` (a core-initiated operation), and `onImageSnapped` from `snapImage()`. But it does NOT fire `onPropertyChanged` from `setProperty()`... 

and that's the backstory for why we have this complicated `_property_change_emission_ensured` situation in pymmcore-plus, (which is receiving an update in this PR to deal with v12 async callbacks).

There are three possible scenarios:

1. Device adapter does NOT fire onPropertyChanged  (most DemoCamera properties, many real devices)
    - Without our override: there's just silence. GUI doesn't update.  this is the main thing we're trying to solve.
2. Device-initiated change (hardware button press, temperature change)
    - No pymmcore-plus API call involved. Only the relay fires. This always worked and still does.
4. Device adapter DOES fire OnPropertyChanged (some real hardware, Core properties via setConfig)
    - this is the hard case and we could *either* just do the simple thing and just let double-fire happen, or do what we're currently doing, and try to ensure only a single onPropertyChanged signal is emitted.  That gets slightly trickier in v12 since the adapter's callback arrives later on a different thread.   So, the solution here is to store a `(dev, prop, val)` token with a (2 second) time to live, which dedupes async calls

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
     "numpy >=1.26.0; python_version >= '3.12'",
     "numpy >=1.25.2",
     "psygnal >=0.10",
-    "pymmcore >=12.0.0.75.0rc1",
+    "pymmcore >=11.10.0.74.0",
     "typing-extensions >=4",                    # not actually required at runtime
     "useq-schema >=0.7.2",
     # cli requirements included by default for now
@@ -62,7 +62,7 @@ PyQt5 = ["PyQt5 >=5.15.4"]
 PyQt6 = ["PyQt6 >=6.4.2"]
 
 [dependency-groups]
-nano = ["pymmcore-nano >=12.0.0.75.0rc1"]
+nano = ["pymmcore-nano==11.10.0.74.0"]
 docs = [
     "mkdocs >=1.4",
     "mkdocs-material>=9.5",


tlambert03:
see https://github.com/pymmcore-plus/pymmcore-widgets/pull/495 for fixes to widgets tests

Issue #514, opened 2025-10-10
Unicore camera class with multi-camera adapter?


dpshepherd:
We've recently setup a four camera darkfield microscope using Balser Ace2 cameras. For this setup, we don't need to control any other hardware. The current MM driver doesn't support the correct version of the Balser pylon SDK for Ace2 cameras, so we made a Unicore camera class using newest Balser pylon SDK / pypylon from the Unicore example and some LLM help. To start, we are using a custom GUI with pymmcore-widgets and one `mmcore` per camera to display and record data.  This strategy works great and serves our immediate needs, so thanks for that! 

While what we did was the fastest way to getting a working instrument, we were curious if it's possible to use the built in multi-camera adapter with Unicore cameras? That would allow us to instead use pymmcore-gui.

Whenever we've used the multi-camera adapter before, we always set it up through the hardware manager by first adding cameras, the multi-camera adapter, and creating groups/properties to do what we needed. It wasn't immediately clear to me how to get that running with a Unicore camera.

Thanks!

dpshepherd:
Coming back around to this. We had tested the PR and decided to keep four independent cores for a few reasons, primarily the need to `startSequenceAcquisition` on the three follower cameras first (they all trigger off the main camera) before we executed `startSequenceAcquisition` on the main camera.

Working with Nikta (the student who built the instrument), we swapped everything over to [ome-writers](https://github.com/pymmcore-plus/ome-writers) this week (thanks again!!). Here, it also still seemed the most logic to keep four separate cores and build on the [pymmcore-plus example](https://github.com/pymmcore-plus/ome-writers/blob/main/examples/pymmcore_plus_example.py) that hooks the `frameReady` callback from an MDA into the ome-writer stream append.

So far, that seems to be working quite well and simplified our control code a lot. It did ping my memory about this discussion because we had to override the `getROI()` function to get the MDAs running. I'm not sure if you want to push on MDAs with multiple Unicore cameras right now, but I'd be happy to think about what an engine pattern for that would look like if so.

PR #609, opened 2026-04-04
refactor: major refactor on UniMMCore pattern, true unification of python devices in C++ core

tlambert03:
This PR is the pymmcore-plus side companion to https://github.com/pymmcore-plus/pymmcore-nano/pull/86

Currently, the UniMMCore pattern essentially maintains TWO cores, a python side one and a c++ side one.  This works as a proof of principle, but it's fundamentally fragile:
- There will always be issues like #514, where the C++ devices don't know about the python devices.
- There will likely be issues like #606 for a long time, where we miss some little synchronization detail between the two cores.
- The current pattern makes it unclear how applications will discover python devices (so, integration in any GUI will be more complicated and special cased).

For these, and many other reasons, the _real_ fix is to have the C++ core *directly* able to control python devices.  And a critical piece of machinery that sort of changes the game here is the `loadMockDeviceAdapter` function added by @marktsuchida in https://github.com/micro-manager/mmCoreAndDevices/pull/648

While it wasn't directly intended for this (and says as much in the docstring), `loadMockDeviceAdapter` is essentially a "hook" into the MMCore plugin system without requiring a compiled shared library.  which is basically exactly what we need to do this the right way.  After that PR, the main missing link is essentially a bridge object: a piece of C++/python binding code that controls a python-side object, but presents a competent MM::Device interface to the C++ side.  That's what https://github.com/pymmcore-plus/pymmcore-nano/pull/86 is

**This would mean that UniMMCore is only supported via pymmcore-nano...** (but I think both @marktsuchida and I agree that ultimately that's the more powerful/flexible binding, and the one we want to use in the long run.  We've been running tests here against nano for a year or two, and it's hard to find any cases anymore where it's not equally good if not better than pymmcore).

This PR removes 3200+ lines of code (all the overrides in _unicore.py where we checked "if python device do one thing, otherwise call super()").  But, importantly, it makes very few changes to the tests (mostly just type coercion between `list/tuple` and `string/float` and some changes to the exceptions that are expected).  So, the existing developer experience when writing python side devices is basically unchanged; in other words: we *don't* need to make python devices as "ugly" as the C++ api, we can have mixins that translate the modern/clean device ABCs and implement the verbose c++ methods under the hood)


tlambert03:
the regression is real on the camera performance:

at 2048 x 2048 uint16: 443 µs (this branch) vs 282 µs (main) ... which is 2.2K fps vs 3.5K fps.   This is because the pure python sequence buffer was able to eliminate an extra copy by passing the device adapter a writeable view directly into the ring buffer.  That's the same feature that could eventually hit main on mmcore (https://github.com/micro-manager/mmCoreAndDevices/pull/570) ... but i think work remains to actually make that useable/safe to merge.  

In any case, I think we're still better than most actual real-world cameras.  For multi-cam we _do_ go back to the difficulties of the single shared circular buffer in mmcore (i.e. restrictions on frame size), but we gain all the advantages as well (standard path, multi-cam device, you could even have 1 python and 1 c++ device managed by MultiCam)

PR #613, opened 2026-04-09
feat: Add optional rich logging via PYMM_LOG_RICH env var

ieivanov:
I actually enjoyed the rich logging :D Would you be open to bringing it back as an optional feature via en environmental variable?

Re-introduce rich logging support as an opt-in feature controlled by the
PYMM_LOG_RICH environment variable. When set to 1/true/yes and rich is
installed, stderr logging uses RichHandler instead of the plain formatter.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


tlambert03 (approved):
yes absolutely :) sorry for removing!

ieivanov:
Thanks! Added a note that rich formatting comes with some overhead.

PR #614, opened 2026-04-10
refactor: delegate per-device timeout to upstream MMCore #914

hinderling:
Originally proposed a Python-side per-device timeout registry. After [@marktsuchida](https://github.com/marktsuchida)'s feedback that the same logic belongs in C++, [mmCoreAndDevices#914](https://github.com/micro-manager/mmCoreAndDevices/pull/914) (mirrored in [mmcore@f003c35](https://github.com/micro-manager/mmcore/commit/f003c35)) shipped the upstream API. This PR is now a thin pymmcore-plus shim over it.

## Motivation

Different devices have very different expected response times, but `waitForDevice` / `waitForSystem` previously applied a single global timeout (`getTimeoutMs()`, default 5 s) to all of them. A motorised XY stage may legitimately need 10+ seconds for a long travel, while a filter wheel change should complete in <200 ms. If the filter wheel is stuck, you'd rather know in 200 ms than wait out the full global timeout.

## API

```python
core.setDeviceTimeoutMs("FilterWheel", 500)    # fail fast
core.setDeviceTimeoutMs("XY", 10_000)          # generous for long travel
core.getDeviceTimeoutMs("FilterWheel")         # → 500 (effective timeout)
core.getDeviceTimeoutMs("UnconfiguredDevice")  # → 5000 (global, no override)
core.hasDeviceTimeout("FilterWheel")           # → True (override registered?)
core.unsetDeviceTimeout("FilterWheel")         # clear → back to global

core.waitForSystem()  # FilterWheel gets 500 ms, XY gets 10 s, others get global
```

The per-call ``timeout_ms`` kwarg from the original proposal has been removed — set/unset is the single mechanism. Timeouts must be positive (`0`/negative raise `RuntimeError`); callers wanting fast-fail-once should call `deviceBusy()` directly.

## What pymmcore-plus does on top of #914

`CMMCorePlus` inherits the four-method API from `pymmcore.CMMCore` verbatim — no Pythonic wrappers. `UniMMCore` overrides the four methods to dispatch by label: Python-device labels use a unicore-local dict (C++ has no record of them); C++ labels delegate to `super()`. This is the only addition strictly necessary for mixed-device support.

## Note on aggregate waits ([mmCoreAndDevices#913](https://github.com/micro-manager/mmCoreAndDevices/issues/913))

`waitForSystem()` / `waitForConfig()` / `waitForDeviceType()` in C++ still loop `waitForDevice()` per device, so wall time can compound across busy devices. The per-device registry is honoured per call, but issue #913 (still open upstream) tracks deadlining the aggregate. We rely on `super().waitForSystem()` for C++ devices and preserve the existing parallel-wait behaviour for Python devices via `ThreadPoolExecutor` in `UniMMCore`.

## TODO before merge

**This PR is blocked on a pymmcore release that includes #914.** Until then, CI fails because the new C++ methods (`setDeviceTimeoutMs` / `getDeviceTimeoutMs` / `hasDeviceTimeout` / `unsetDeviceTimeout`) don't exist on the released `pymmcore` (≤12.2.2). Local testing was done against a source build of `pymmcore@main` with `subprojects/mmcore.wrap` repointed to a post-#914 mmcore commit (currently `8399a00`).

- [ ] Wait for a `pymmcore` release that ships #914.
- [ ] **Bump `pymmcore` minimum version in `pyproject.toml`** to that release. Currently allows `pymmcore>=11.10`.
- [ ] Drop the `# type: ignore[misc]` comments on `super().{set,get,has,unset}DeviceTimeout(...)` calls in `UniMMCore` once pymmcore stubs ship #914.

## Test plan

- [x] Registry CRUD: `setDeviceTimeoutMs` / `getDeviceTimeoutMs` (returns effective timeout) / `hasDeviceTimeout` / `unsetDeviceTimeout`
- [x] "Must be positive" rejects 0 and negative
- [x] Unknown-device labels raise `RuntimeError`
- [x] Global `setTimeoutMs` is independent from per-device registry
- [x] `UniMMCore` dispatches by label; per-device registry honoured for Python devices
- [x] `UniMMCore.waitForSystem` honours per-device registry for Python devices and propagates `TimeoutError`
- [x] Full local suite: 719 passed, 5 skipped, 1 xfailed against `pymmcore@main + mmcore@HEAD`

### DIFF
--- a/src/pymmcore_plus/core/_mmcore_plus.py
+++ b/src/pymmcore_plus/core/_mmcore_plus.py
@@ -261,6 +261,9 @@ class CMMCorePlus(pymmcore.CMMCore):
         self._last_config: tuple[str, str] = ("", "")
         # last position set by setXYPosition, None means currentXYStageDevice
         self._last_xy_position: dict[str | None, tuple[float, float]] = {}
+        # per-device timeout overrides (label → ms).  Consulted by
+        # waitForDevice / waitForSystem when no per-call timeout_ms is given.
+        self._device_timeouts: dict[str, float] = {}
 
         self._mm_path = mm_path or find_micromanager()
         if not adapter_paths and self._mm_path:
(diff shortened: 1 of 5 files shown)


tlambert03:
interesting feature.  I can see the potential use case.  I kinda want to hear @marktsuchida's thoughts here.  (including whether it should be a C++ feature, or, if not, whether it's "safe" from a maintenance, functionality, and scope creep perspective, on the pymmcore-plus side).  It does seem like something that would be hard/annoying for an end-user to do without subclassing

ieivanov:
I agree this would be a useful feature. We currently keep our core timeout very long - 5-10 seconds, because the XY stages takes a while to move between far-away distances. 

As a robustness feature we also have a retry mechanism where we catch timeout errors and re-issue the `setXYStagePosition` command call several times before giving up.

marktsuchida:
> whether it should be a C++ feature

It certainly could be, and that might be desirable because then it would apply to `waitForSystem()` and `waitForConfig()` calls naturally. Also, even if not immediately easy to do, it would make sense for this kind of setting to go in the hardware config file in the future.

I'll see if I can create a PR in mmCoreAndDevices.

marktsuchida:
Looking at the C++ code I see that the "timeout" is not handled in a logical way for the aggregate wait functions (`waitForSystem()`, `waitForConfig()`, etc.) — currently, it just loops over calls to `waitForDevice()`, which means the effective timeout is indeterminate (https://github.com/micro-manager/mmCoreAndDevices/issues/913).

It will be good to change that, so it's probably good to avoid duplicating that behavior in pymmcore-plus.

tlambert03:
Thanks for weighing in @marktsuchida!  Will keep an eye on the c++ discussion and update here

hinderling:
updated to reflect the implementation in https://github.com/micro-manager/mmCoreAndDevices/pull/914 

> As a robustness feature we also have a retry mechanism where we catch timeout errors and re-issue the `setXYStagePosition` command call several times before giving up.

+1 for this feature, we also have custom workarounds for this for several unreliable devices

hinderling:
Note: even if https://github.com/micro-manager/mmCoreAndDevices/issues/913 is solved, mixed unicore/mmcore configs would still be indeterminant due to sequential wait calls to mmcore/unicore. Would require concurrent C++/Python wait phases.

tlambert03:
> mixed unicore/mmcore configs would still be indeterminant due to sequential wait calls to mmcore/unicore. Would require concurrent C++/Python wait phases.

i believe that would be fixed by the big unicore refactor that gives the C++ side control/awareness of the python devices?  

https://github.com/pymmcore-plus/pymmcore-plus/pull/609

_edit: actually this whole thing will probably be unnecessary after that_

PR #617, opened 2026-04-22
feat: Modify buffer overflow error message

ieivanov:
Closes #611


tlambert03 (commented):
If we're going to list APIs let's mention the primary one, setCircularBufferSize (or whatever it is... not at computer at the moment).  It's ok to also mention the pymmcore-plus specific env var: but this is an old concept with lower level APIs that we should highlight)

ieivanov:
Good point, fixed!

tlambert03 (approved):

PR #618, opened 2026-05-06
feat: unified logging via MMCore

marktsuchida:
Closes #385.

This uses https://github.com/micro-manager/mmCoreAndDevices/pull/906 and https://github.com/micro-manager/mmCoreAndDevices/pull/909 to send all Python pymmcore-plus logging to `CMMCore.log()`, with log sinks and rotation handled by MMCore. We then no longer have the situation of two things writing to the same file; ordering is guaranteed.

`PYMM_LOG_RICH` is removed because there is no Python-side formatting any more (with apologies to @ieivanov #613). (MMCore now has a tiny bit of coloring for stderr but of course nothing like Rich. Not sure if there is anything that can be done here.)

Questions remain about what happens when there is more than one CMMCore, but I think that was already semi-broken?

If anything is logged to the `pymmcore-plus` logger before a `CMMCore` is created, it will use Python's defaults. I suppose we could try to preserve such log entries, but I don't think it's worth the trouble given typical use patterns.

`configure_logging()` can be called before creating a `CMMCore`, in which case its effects will only be seen once the `CMMCore` is created.

Bumps the required pymmcore version.

I'm leaving this as draft for now because I want to go over once more especially the docs and tests. @tlambert03, let me know if the general direction looks good.

PR #620, opened 2026-05-14
fix: close handlers and snapshot list in configure_logging cleanup

hinderling:
The handler-cleanup loop in configure_logging() had two intra-process bugs that surface when configure_logging() runs more than once in the same interpreter (uncommon in production, common in tests, notebooks with %autoreload, and on explicit reconfig calls):

1. Mutate-during-iteration: removeHandler mutates logger.handlers, so `for handler in logger.handlers: logger.removeHandler(handler)` skips every other element. With the typical [stderr, file] handler pair, the file handler is never actually removed, and the next call to configure_logging() stacks a fresh file handler on top of it. Every log message then writes twice to the file (verified locally).

2. Missing close(): handlers that do get removed keep their underlying file descriptors open. On POSIX rename-while-open is fine; on Windows the orphan FD can lock the rotation target so doRollover fails with PermissionError on the next rotation attempt.

Iterate a list snapshot and call handler.close() after detaching.

This does not address the cross-process case where multiple Python interpreters share the default log path - that needs per-PID log files (separate issue).

### DIFF
--- a/src/pymmcore_plus/_logger.py
+++ b/src/pymmcore_plus/_logger.py
@@ -116,8 +116,9 @@ def configure_logging(
     root_logger = logging.getLogger()
     root_logger.setLevel(logging.DEBUG)
 
-    for handler in logger.handlers:
+    for handler in list(logger.handlers):
         logger.removeHandler(handler)
+        handler.close()
 
     # automatically log to stderr
     if log_to_stderr and sys.stderr:


tlambert03:
i suspect https://github.com/pymmcore-plus/pymmcore-plus/pull/618 makes this obsolete, (and probably also fixes cross-process case).  but this is a minor/safe enough change anyway so let's just merge

PR #629, opened 2026-09-14
fix(unicore): refresh cached State/Label when a Python state device moves

hinderling:
### Problem

`UniMMCore` keeps a Python-side property cache (`_state_cache`) for Python devices. `getProperty` and `setProperty` maintain it, but `setState` and `setStateLabel` move the device without touching it, and `setProperty(dev, "State", ...)` leaves `Label` stale (and vice versa). CMMCore refreshes both keywords whenever a state device moves.

The stale entries surface through every cached read: `getPropertyFromCache`, `getSystemStateCache`, and `getCurrentPixelSizeConfig(cached=True)`. The last one matters most: `summary_metadata` calls `getPixelSizeUm(True)` and `getPixelSizeAffine(True)`, so after switching a Python objective turret with `setStateLabel`, every MDA records the *previous* objective's pixel size. It only heals once something reads the property live.

Minimal repro on a Python `StateDevice` with labels:

```python
core.setStateLabel("Objective", "60x")
core.setStateLabel("Objective", "10x")
core.getPropertyFromCache("Objective", "Label")   # "60x" on main
```

### Fix

A small `_cache_state_props` helper re-reads `State` and `Label` from the device and stores both, called from `setState`, `setStateLabel`, and from `setProperty` when the property is `State` or `Label` on a `StateDevice`. Values are cached in the same raw form `getProperty` uses.

Test: `test_cached_state_and_label_follow_the_device`, parametrized over a Python and a C++ state device like the rest of `test_state.py`. The C++ variant passes on main and documents the CMMCore behaviour; the Python variant fails on main and passes here.

### Relation to #609

#609 removes the dual-core layer, including this cache, and will fix this at the root.


tlambert03 (approved):
thanks!

PR #630, opened 2026-09-14
fix: rotate the log file while MMCore writes to it

ieivanov:
## Problem

`CMMCorePlus.__init__` points the C++ core's CoreLog at the same file the Python `RotatingFileHandler` writes to (`setPrimaryLogFile(current_logfile(logger))`, next to a `# TODO: test this on windows` comment). The core keeps its own handle on that file, and on Windows a file cannot be renamed while another handle is open. So once `pymmcore-plus.log` reaches `maxBytes`, every rollover fails:

```
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process:
  '...\pymmcore-plus.log' -> '...\pymmcore-plus.log.1'
```

The stdlib handler drops the record, prints a traceback to stderr, and tries (and fails) again on the very next record, for the rest of the session. In long acquisitions this happens after a few hours; roughly three quarters of the file's lines come from the C++ core, so the limit is reached quickly. Only workaround so far was to stop everything and rename the file by hand.

## Fix

- `_logger._RotatingFileHandler`: releases the registered cores' primary log file (`setPrimaryLogFile("")`) before the rename and points them at the fresh file afterwards. 
- `_logger.follow_logfile(core)`: registers a core (weak reference) and points its primary log file at the current pymmcore-plus log file. `CMMCorePlus.__init__` uses it instead of calling `setPrimaryLogFile` directly. Registered cores also follow a later `configure_logging()` call (new file or `file=None`).
- If a rollover still fails (for example a second pymmcore-plus process shares the default log file), the handler keeps appending to the current file, writes one note into the log, and retries after `rollover_retry_interval` (60 s) instead of dropping records and spamming stderr.

Behaviour is otherwise unchanged: Python and core messages still interleave in one file, `current_logfile()` still returns it, and `core.getPrimaryLogFile()` still points at it.

## Tests

`tests/test_logger.py`:
- rotation with a live core succeeds and the core continues writing into the fresh file (fails on Windows without this change)
- cores follow `configure_logging()` to a new file and release the old one
- a still-failing rollover keeps every record, notes the failure once, backs off, and rotates once the file is free

Also verified end to end on Windows with a demo-config MDA and `maxBytes` forced to 8 kB: eight rollovers mid-acquisition, empty stderr, core and Python lines present in every rotated file.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ieivanov:
Ah, yes, good catch, I had forgotten about #618. Up to you if this is worth the "emergency" patch or if we can wait for #618 to merge with a cleaner resolution. I suspect this issue has been around for a while, and I'm guessing other users have learned to troubleshoot it the way I have. Do you know if #618 can be merged soon? Happy to run a few tests of the new behavior if that would be useful.

PR #631, opened 2026-09-14
fix: don't read state-device properties the device doesn't have

ieivanov:
## Problem

`CMMCorePlus._property_change_emission_ensured` expands a single `State` or `Label` request on a state device into both properties, then reads each one before and after the operation:

```python
if (
    len(properties) == 1
    and properties[0] in STATE_PROPS
    and self.getDeviceType(device) is DeviceType.StateDevice
):
    properties = STATE_PROPS
try:
    before = [self.getProperty(device, p) for p in properties]
except Exception as e:
    logger.warning(...)
    yield
    return
```

`Label` is conventional for state devices, but not mandatory — some adapters implement only `State`. The one that surfaced this for us is the TriggerScopeMM TTL switch (`CStateDeviceBase`, `State` + one property per TTL line, no `Label`).

For such a device, every `setProperty(dev, "State", ...)`, `setState` and `setStateLabel`:

1. asks MMCore for `Label`, which raises and logs an error on the device logger;
2. logs a WARNING here;
3. falls into the `yield; return` path, so **no `propertyChanged` signal is emitted at all**.

During a hardware-sequenced MDA that's once per event — the core log fills with

```
[ERR,Core:dev:TS2_TTL1-8] Cannot get value of property "Label" [ Error in device "TS2_TTL1-8": Invalid property name encountered: Label (2) ]
```

and the `State` change the context manager exists to guarantee is the one thing that doesn't get reported.

## Change

Filter the monitored properties through `hasProperty` before reading them, so only properties the device actually implements are queried. `State` events are emitted normally on devices that lack `Label`, and nothing is logged.

The filter is applied only when `State`/`Label` are involved, so an explicit `setProperty` with a genuinely bogus property name still takes the existing warning path rather than being silently ignored.

## Testing

New `tests/test_events.py::test_set_state_events_without_label_property` patches `hasProperty`/`getProperty` on the demo `Objective` to emulate a state device with no `Label`, and asserts that `Label` is never requested, nothing is logged, and the `State` change is still emitted. Confirmed it fails on `main` with exactly the warning above and passes with the fix; `tests/test_events.py` and `tests/test_core.py` pass, ruff clean.

Marked draft: happy to adjust where the filtering goes, e.g. applying it to all monitored properties rather than just the state ones.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ieivanov (commented):
Confirmed with TriggerScope hardware that this change eliminates the `Cannot get value of property "Label"` noise in the logs.

tlambert03:
Thanks @ieivanov, nice find.  `Label` is indeed optional (MMCore guards it with `HasProperty` too). I'd go with your alternative though: let's just filter everything through `hasProperty` and drop the `any(... in STATE_PROPS)` special case. A bogus property name would still raise in `super().setProperty()` anyway, so we're not silently ignoring anything, and the pre-read warning is just extra noise in that case. Then happy to merge 🙏

ieivanov:
Done! Thanks for the suggestion

